PR #7, opened 2025-04-14
CUDA arch independent

EduardDurech:
CUDA dir was hardcoded

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,15 +24,41 @@ def _find_cuda_home():
             cuda_home = '/usr/local/cuda'
     return cuda_home
 
+def _find_cuda_dir(cuda_home, dir_name, target_file=None, alternative_dir=None, subdir=None):
+    """Find CUDA directories."""
+    cuda_home = Path(cuda_home)
+    dirs = []
 
-cuda_home = Path(_find_cuda_home())
-include_dirs = [
-    str(cuda_home.resolve() / 'targets/x86_64-linux/include'),
-]
-library_dirs = [
-    str(cuda_home.resolve() / 'lib64'),
-    str(cuda_home.resolve() / 'lib64/stubs'),
-]
+    # Check if main directory exists
+    if (cuda_dir := cuda_home / dir_name).is_dir():
+        dirs.append(str(cuda_dir.resolve()))
+    # Check alternative directory paths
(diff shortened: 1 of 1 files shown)


fzyzcjy:
Good job!

@ocss884 (who wrote this part of code) could you please have a look and check whether that works on your env? Thanks!

ocss884:
LGTM, @EduardDurech I'm just curious is there any cases the hardcoded path doesn' exists? I assume the file structure of CUDA toolkit is somehow stable.

EduardDurech:
@ocss884 more archs, e.g. if aarch64 the x86_64 wouldn't work, in terms of searching for the C headers it's just backup

fzyzcjy:
OK so if the code works for both @ocss884 and @EduardDurech's env then it looks OK

setup.py
```diff
@@ -24,9 +24,17 @@ def _find_cuda_home():
...
+include_dirs = []
+if (cuda_home / 'include').is_dir():
+    include_dirs.append(str((cuda_home / 'include').resolve()))
+else:
+    # Search for 'cuda_runtime_api.h' in subdirectories
+    for path in cuda_home.rglob('cuda_runtime_api.h'):
+        include_dirs.append(str(path.parent.resolve()))
+        break
+    else:
+        raise RuntimeError("Could not find CUDA include directory.")
```
fzyzcjy: nit: what about extracting it to a function, say `def find_include_dirs()`

EduardDurech: It can be abstracted, I won't be able to test until next week but feel free to make any edits you want, else will do later

fzyzcjy: Take your time! I am a bit busy now so may or may not have time to extract this and test locally recently.

ocss884:
> OK so if the code works for both @ocss884 and @EduardDurech's env then it looks OK

No problem on my machine

setup.py
EduardDurech: @fzyzcjy updated for `include_dirs` and `library_dirs` from #11

fzyzcjy: Good job! Too busy now but I hope to merge soon

EduardDurech:
Replaced by https://github.com/fzyzcjy/torch_memory_saver/pull/19

PR #8, opened 2025-04-14
`LD_PRELOAD` Script

EduardDurech:
- Package data contained in torch_memory_saver
- tms python sets `LD_PRELOAD`

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,9 +19,43 @@ memory_saver.pause()
 memory_saver.resume()
 ```
 
+## Installation
+
+For the library to work properly, it needs to be preloaded using the `LD_PRELOAD` environment variable.
+
+### Option 1: Run a single command
+
+After installation, you can run your scripts with:
+
+```bash
+tms python your_script.py
+```
+
+### Option 2: Activate for the current shell session
+
+To activate torch_memory_saver for all Python programs in your current shell:
+
+```bash
+source tms
+# Now run your Python programs normally
+python your_script.py
+```
+
+### Option 3: Add to your shell profile
+
+To make torch_memory_saver available in all your shells, add this line to your `~/.bashrc` or `~/.zshrc`:
+
+```bash
+source $(which tms)
+```
+
+## Technical Details
+
+This library uses CUDA driver API functions to manage the lifetime of CUDA memory allocations. It requires `LD_PRELOAD` to intercept CUDA memory allocation calls made by PyTorch.
+
 Please refer to https://github.com/sgl-project/sglang/issues/2542#issuecomment-2563641647 for details.
 
-TODO:
+## TODO:
 
 - [x] Implementation
 - [x] Publish to pypi
--- a/scripts/tms
+++ b/scripts/tms
@@ -1,0 +1,39 @@
+#!/bin/bash
+
+# Get the absolute path to the library
+LIB_PATH=$(python -c "
+import torch_memory_saver, importlib.machinery, pathlib, os
+package_dir = pathlib.Path(torch_memory_saver.__file__).parent
+ext_suffix = importlib.machinery.EXTENSION_SUFFIXES[0]
+lib_name = f'torch_memory_saver_cpp{ext_suffix}'
+print(str(package_dir / lib_name))
+")
+
+# Detect if script is being sourced
+(return 0 2>/dev/null) && SOURCED=1 || SOURCED=0
+
+# Handle sourcing case
+if [ $SOURCED -eq 1 ]; then
+    # When the script is sourced, we export LD_PRELOAD and return
+    export LD_PRELOAD="$LIB_PATH${LD_PRELOAD:+:$LD_PRELOAD}"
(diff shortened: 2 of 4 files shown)


scripts/tms
fzyzcjy, line 1: nit: what about naming `tms` to `torch_memory_saver`

scripts/tms
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+
+# Get the absolute path to the library
+LIB_PATH=$(python -c "
+import torch_memory_saver, importlib.machinery, pathlib, os
+package_dir = pathlib.Path(torch_memory_saver.__file__).parent
+ext_suffix = importlib.machinery.EXTENSION_SUFFIXES[0]
+lib_name = f'torch_memory_saver_cpp{ext_suffix}'
+print(str(package_dir / lib_name))
```
fzyzcjy, line 9: nit: what about we provide such a function in torch_memory_saver python code, and here we only need to do as simple as `$(python -c "import torch_memory_saver; torch_memory_saver.that_function()"`

scripts/tms
```diff
@@ -0,0 +1,39 @@
...
+")
+
+# Detect if script is being sourced
+(return 0 2>/dev/null) && SOURCED=1 || SOURCED=0
```
fzyzcjy, line 13: nit: curious whether this will accurately detect whether it is being sourced, or will there false positive / negative?

torch_memory_saver/__init__.py
```diff
@@ -59,7 +58,10 @@ def enabled(self):
     def compute():
         env_ld_preload = os.environ.get('LD_PRELOAD', '')
         if 'torch_memory_saver' in env_ld_preload:
-            return _BinaryInfo(cdll=ctypes.CDLL(env_ld_preload))
+            paths = env_ld_preload.split(':')
+            lib_path = next((path for path in paths if 'torch_memory_saver' in path), None)
+            if lib_path:
```
fzyzcjy, line 63: nit: looks like we may have a bug for the else case (or, if the `else` branch will never occur, we may change if to assert)

torch_memory_saver/__init__.py
```diff
@@ -82,34 +84,4 @@ def next_id(self):
...
-    old_value = os.environ.get(key, '')
-    os.environ[key] = value
-    logger.debug(f'change_env set key={key} value={value}')
-    try:
-        yield
-    finally:
-        assert os.environ[key] == value
-        os.environ[key] = old_value
-        logger.debug(f'change_env restore key={key} value={old_value}')
+_global_info = _GlobalInfo()
```
fzyzcjy, line 87: well, it would be great to keep the old configure_subprocess etc, because it is how SGLang uses it. https://github.com/sgl-project/sglang/blob/8aab7fdb21e86b87a3141951ce498246fd4aae73/python/sglang/srt/entrypoints/engine.py#L522

README.md
```diff
@@ -19,9 +19,43 @@ memory_saver.pause()
 memory_saver.resume()
 ```
 
+## Installation
+
+For the library to work properly, it needs to be preloaded using the `LD_PRELOAD` environment variable.
```
fzyzcjy, line 24: nit: would be great to mention the SGLang approach, which indeed does not require any operations

fzyzcjy (commented):
Great job!

fzyzcjy:
@EduardDurech Btw, I am curious how/where are you using torch_memory_saver?

EduardDurech:
@fzyzcjy right now standalone API

```python
import os, torch
from torch_memory_saver import TorchMemorySaver

memory_saver = TorchMemorySaver()
print(os.system('nvidia-smi'))

with memory_saver.region():
    x = torch.full((1_000_000_000,), 100, dtype=torch.uint8, device='cuda')

print(os.system('nvidia-smi'))

memory_saver.pause()
print(os.system('nvidia-smi'))

memory_saver.resume()
print(os.system('nvidia-smi'))
```

Testing, will use with veRL-SGLang, did you guys implement `LD_PRELOAD` there with subprocess?

fzyzcjy:
> will use with veRL-SGLang, did you guys implement LD_PRELOAD there with subprocess?

Yes, if you are using memory saver inside SGLang, then the LD_PRELOAD is automatically configured and you do not need to do anything .

EduardDurech:
Okay, do you know where in the code it finds the .so? I couldn't find invocation of torch_memory_saver, maybe it would be good to be compatible as standalone and as dependency, but currently the PR moves the .so to packages/torch_memory_saver instead of packages

fzyzcjy:
> do you know where in the code it finds the .so? 

Just follow the configure_subprocess function you will see

torch_memory_saver/__init__.py
EduardDurech: Will look at the approach later and modify PR, not by system for a few days

fzyzcjy: Take your time!

PR #17, opened 2025-06-06
Support Mulitple Instance

hebiao064:
## Why

vLLM has supported Multi-Stage Awake for vllm engine: https://github.com/vllm-project/vllm/issues/15254

But in SGLang, we are using `torch_memory_saver` for holding the model weight and KV Cache virtual address (to make sure CUDA Graph works across different rollouts)

And `torch_memory_saver` is a singleton, which make it hard for SGLang to support Multi-Stage Awake which is critical in RL use case

## More context:
https://github.com/vllm-project/vllm/issues/15254 
https://github.com/sgl-project/sglang/blob/main/python/sglang/srt/managers/scheduler.py#L2111 
https://github.com/sgl-project/sglang/blob/27e327b415cfd009ce1cf747b3fc0a1bbe7ee15b/test/srt/test_release_memory_occupation.py#L21 
https://github.com/fzyzcjy/torch_memory_saver?tab=readme-ov-file 
https://github.com/sgl-project/sglang/issues/2542#issuecomment-2563641647 
https://github.com/sgl-project/sglang/issues/2583 
https://github.com/vllm-project/vllm/blob/main/vllm/worker/worker.py#L145 
https://github.com/sgl-project/sglang/issues/6367

### DIFF
--- a/csrc/torch_memory_saver.cpp
+++ b/csrc/torch_memory_saver.cpp
@@ -136,7 +136,7 @@ public:
         CUDAUtils::cu_mem_set_access(*ptr, size, device);
 
         {
-            const std::lock_guard <std::mutex> lock(allocator_metadata_mutex_);
+            const std::lock_guard<std::mutex> lock(allocator_metadata_mutex_);
             allocation_metadata_.emplace(*ptr, _AllocationMetadata{size, device, allocHandle});
         }
 
--- a/README.md
+++ b/README.md
@@ -27,3 +27,17 @@ TODO:
 - [x] Publish to pypi
 - [ ] More tests and infra
 - [ ] Documentation
+
+
+## Development
+
+```bash
+pip install -e .
+```
+
+An `torch_memory_saver_cpp.abi3.so` will be built under `build/` folder.
+
+You can use this command for testing:
+```bash
+LD_PRELOAD=/home/jobuser/torch_memory_saver/torch_memory_saver_cpp.abi3.so python examples/simple.py
+```
--- a/examples/simple.py
+++ b/examples/simple.py
@@ -8,12 +8,19 @@ logging.basicConfig(level=logging.DEBUG, stream=sys.stdout)
 
 from torch_memory_saver import TorchMemorySaver
 
+import os
+
 memory_saver = TorchMemorySaver()
 
-normal_tensor = torch.full((1_000_000_000,), 100, dtype=torch.uint8, device='cuda')
+normal_tensor = torch.full((1_000_000,), 100, dtype=torch.uint8, device='cuda')
 
 with memory_saver.region():
-    pauseable_tensor = torch.full((1_000_000_000,), 100, dtype=torch.uint8, device='cuda')
+    pauseable_tensor = torch.full((10_000_000_000,), 100, dtype=torch.uint8, device='cuda')
+
+# Get the virtual address
+original_address = pauseable_tensor.data_ptr()
+print(f"Tensor virtual address: 0x{original_address:x}")
+os.system("nvidia-smi --query-gpu=memory.used --format=csv,noheader,nounits")
 
 print(f'{normal_tensor=} {pauseable_tensor=}')
 
@@ -21,11 +28,18 @@ print('sleep...')
 time.sleep(3)
 
 memory_saver.pause()
+os.system("nvidia-smi --query-gpu=memory.used --format=csv,noheader,nounits")
 
 print('sleep...')
 time.sleep(3)
 
 memory_saver.resume()
+os.system("nvidia-smi --query-gpu=memory.used --format=csv,noheader,nounits")
+
+new_address = pauseable_tensor.data_ptr()
+print(f"Tensor virtual address: 0x{new_address:x}")
+
+assert original_address == new_address, 'Tensor virtual address should be the same'
 
 print('sleep...')
 time.sleep(3)
--- a/setup.py
+++ b/setup.py
@@ -31,7 +31,7 @@ def _get_platform_architecture():
         try:
             uname_output = subprocess.check_output(["uname", "-a"], encoding="utf-8")
             if "tegra" in uname_output:
-                return f"{"tegra"}-{host_arch}"
+                return f"tegra-{host_arch}"
         except Exception as e:
             print(f"[warn] Failed to run uname: {e}")
 


fzyzcjy:
Great job! Yes that should be implemented but I am too busy with SGLang... I will review and (if that's OK) quickly merge it after monday since have really no time these two days given tight deadlines.

hebiao064:
> Great job! Yes that should be implemented but I am too busy with SGLang... I will review and (if that's OK) quickly merge it after monday since have really no time these two days given tight deadlines.

Sure, I will start working on SGLang integration now, I just need some guidance from the community to make sure this could work cuz this is still new to me. Thanks

csrc/torch_memory_saver.cpp
```diff
@@ -211,78 +212,126 @@ class TorchMemorySaver {
...
+        return current_saver_;
     }
+}
 
-    bool is_interesting_region() {
-        return is_interesting_region_;
+// ----------------------------------------------- default instance ------------------------------------------------
+
+namespace {
+    // Provide a default instance for backward compatibility
```
fzyzcjy: no worries for this backward compatibility, since nobody is using our c++ api

torch_memory_saver/__init__.py
```diff
@@ -59,7 +68,19 @@ def enabled(self):
...
-            return _BinaryInfo(cdll=ctypes.CDLL(env_ld_preload))
+            try:
+                cdll = ctypes.CDLL(env_ld_preload)
+                # Define function signatures
+                cdll.tms_create.restype = ctypes.c_void_p
+                cdll.tms_destroy.argtypes = [ctypes.c_void_p]
+                cdll.tms_region_enter.argtypes = [ctypes.c_void_p]
+                cdll.tms_region_leave.argtypes = [ctypes.c_void_p]
+                cdll.tms_pause.argtypes = [ctypes.c_void_p]
+                cdll.tms_resume.argtypes = [ctypes.c_void_p]
```
fzyzcjy: nit: extract to function

fzyzcjy (commented):
very briefly glanced and will more carefully check later, but anyway here are two nits

csrc/torch_memory_saver.cpp
hebiao064: removed, and tested with all examples

torch_memory_saver/__init__.py
hebiao064: done

csrc/torch_memory_saver.cpp
```diff
@@ -211,78 +212,114 @@ class TorchMemorySaver {
...
     } else {
         return APIForwarder::call_real_cuda_free(ptr);
     }
 }
 
 extern "C" {
-void tms_region_enter() {
-    RegionManager::enter();
+void tms_region_enter(TorchMemorySaver* saver) {
+    if (saver == nullptr) {
```
fzyzcjy, line 264: nit: fail fast and just exit(1) for this, since this should never happen

(same for all others)

csrc/torch_memory_saver.cpp
```diff
@@ -211,78 +212,114 @@ class TorchMemorySaver {
...
+        std::cerr << "[torch_memory_saver.cpp] Error: NULL saver passed to tms_region_enter" << std::endl;
+        return;
+    }
+    RegionManager::set_current_saver(saver);
```
fzyzcjy, line 276: nit: add assert that the old saver is null

csrc/torch_memory_saver.cpp
```diff
@@ -211,78 +212,114 @@ class TorchMemorySaver {
...
 
-void tms_region_leave() {
-    RegionManager::leave();
+void tms_region_leave(TorchMemorySaver* saver) {
+    if (saver == nullptr) {
+        std::cerr << "[torch_memory_saver.cpp] Error: NULL saver passed to tms_region_leave" << std::endl;
+        return;
+    }
+    saver->leave_region();
+    RegionManager::set_current_saver(nullptr);
```
fzyzcjy, line 293: nit: add assert that the_current_saver == the_saver_that_is_passed_in

csrc/torch_memory_saver.cpp
```diff
@@ -211,78 +212,114 @@ class TorchMemorySaver {
...
+    }
+    saver->resume();
 }
+
+TorchMemorySaver* tms_create() {
+    return new TorchMemorySaver();
+}
+
+void tms_destroy(TorchMemorySaver* saver) {
+    if (saver != nullptr) {
```
fzyzcjy: nit: just assert saver != nullptr and exit(1) if it happens

csrc/torch_memory_saver.cpp
```diff
@@ -211,78 +212,114 @@ class TorchMemorySaver {
...
 
-    bool is_interesting_region() {
+    bool is_interesting_region() const {
         return is_interesting_region_;
     }
+
+private:
+    std::mutex allocator_metadata_mutex_;
+    std::unordered_map<void *, _AllocationMetadata> allocation_metadata_;
+    bool is_interesting_region_;
```
fzyzcjy: it seems that, when current saver is not none, we are always in interesting region. thus wondering whether we can remove this bool flag

fzyzcjy (commented):
too tired and only briefly checked and here are some nits

csrc/torch_memory_saver.cpp
hebiao064: done

csrc/torch_memory_saver.cpp
hebiao064: done

csrc/torch_memory_saver.cpp
hebiao064: done

csrc/torch_memory_saver.cpp
hebiao064: done

csrc/torch_memory_saver.cpp
hebiao064: done

hebiao064:
> generally LGTM, I will need to find time to check it more in detail later

thanks in advance!

fzyzcjy:
IIRC you said https://github.com/fzyzcjy/torch_memory_saver/pull/20 should be used so I close this one. Feel free to ping me if need to reopen!

PR #20, opened 2025-06-11
[Alternative of Multi Instance Solution] Singleton with tag based resume

hebiao064:


torch_memory_saver/__init__.py
```diff
@@ -55,11 +77,27 @@ class _BinaryInfo:
...
+        cdll.tms_enable.argtypes = []
+        cdll.tms_disable.argtypes = []
+        cdll.tms_set_current_tag.argtypes = [ctypes.c_char_p]
+        cdll.tms_pause.argtypes = [ctypes.c_char_p]
+        cdll.tms_resume.argtypes = [ctypes.c_char_p]
+
     @staticmethod
     def compute():
-        env_ld_preload = os.environ.get('LD_PRELOAD', '')
+        env_ld_preload = "/home/jobuser/torch_memory_saver/torch_memory_saver_cpp.abi3.so"
```
hebiao064: need to revert this line, this is for internal testing

torch_memory_saver/__init__.py
```diff
@@ -96,13 +132,13 @@ def get_binary_path():
...
     ]
-    assert len(candidates) == 1, f'{candidates=}'
+    assert len(candidates) == 1, f'Expected exactly one torch_memory_saver_cpp library, found: {candidates}'
     return candidates[0]
 
 
 @contextmanager
 def configure_subprocess():
-    with change_env('LD_PRELOAD', str(get_binary_path())):
+    with change_env('LD_PRELOAD', "/home/jobuser/torch_memory_saver/torch_memory_saver_cpp.abi3.so"):
```
hebiao064: need to revert this line, this is for internal testing

csrc/torch_memory_saver.cpp
```diff
@@ -6,8 +6,9 @@
 #include <dlfcn.h>
 #include <unordered_map>
 #include <mutex>
+#include <string>
 
-// #define TMS_DEBUG_LOG
+#define TMS_DEBUG_LOG
```
hebiao064: need revert it

fzyzcjy:
btw I see CUDA graph in your figure, thus https://github.com/fzyzcjy/torch_memory_saver/issues/21

csrc/torch_memory_saver.cpp
```diff
@@ -210,79 +243,67 @@ class TorchMemorySaver {
...
-        return instance;
-    }
-
 private:
-    // Similar to torch's CUDACachingAllocator and CUDAPluggableAllocator
+    TorchMemorySaver() : enabled_(false), current_tag_("default") {}
+
     std::mutex allocator_metadata_mutex_;
     std::unordered_map<void *, _AllocationMetadata> allocation_metadata_;
+    bool enabled_;
```
fzyzcjy: bug: this seems not threadlocal anymore, thus multi-thread using `with saver.region()` seems to becomes unsafe

csrc/torch_memory_saver.cpp
```diff
@@ -210,79 +243,67 @@ class TorchMemorySaver {
...
+    std::string current_tag_;
```
fzyzcjy: for `enabled_` and `current_tag_`, what about doing something like:

```
class TmsThreadLocalContext {
  bool enabled_;
  std::string tag_;
};

static thread_local TmsThreadLocalContext current_thread_context_;
```

torch_memory_saver/__init__.py
```diff
@@ -14,29 +14,50 @@
 class TorchMemorySaver:
     def __init__(self):
         self._mem_pool = None
-        self._id = _global_info.next_id()
-        assert self._id == 1, 'Only support one single instance yet (multi-instance will be implemented later)'
+        if not _global_info.binary_info.enabled:
```
fzyzcjy: nit: maybe not warn this early, since it is possible users import it and not use it when they do not have the binary

torch_memory_saver/__init__.py
```diff
@@ -14,29 +14,50 @@
...
     @contextmanager
-    def region(self):
+    def region(self, tag: str = "default"):
+        """Context manager for memory saving with optional tag"""
         if _global_info.binary_info.enabled:
             self._ensure_mem_pool()
             with torch.cuda.use_mem_pool(self._mem_pool):
-                _global_info.binary_info.cdll.tms_region_enter()
+                _global_info.binary_info.cdll.tms_set_current_tag(tag.encode('utf-8'))
+                _global_info.binary_info.cdll.tms_enable()
```
fzyzcjy: nit: what about keeping the origin "is_interesting_region" naming to minimize diff

torch_memory_saver/__init__.py
```diff
@@ -14,29 +14,50 @@
...
 
-    def resume(self):
+    def resume(self, tag: Optional[str] = None):
+        """Resume memory for specific tag or all memory if tag is None"""
         if _global_info.binary_info.enabled:
-            _global_info.binary_info.cdll.tms_resume()
+            tag_bytes = tag.encode('utf-8') if tag else None
+            _global_info.binary_info.cdll.tms_resume(tag_bytes)
+
+    def set_current_tag(self, tag: str):
```
fzyzcjy: maybe we do not need to expose this API, since users will directly use the `region` api

torch_memory_saver/__init__.py
```diff
@@ -14,29 +14,50 @@
...
+
+    def enable(self):
+        """Enable memory saving"""
+        if _global_info.binary_info.enabled:
+            _global_info.binary_info.cdll.tms_enable()
+
+    def disable(self):
+        """Disable memory saving"""
+        if _global_info.binary_info.enabled:
+            _global_info.binary_info.cdll.tms_disable()
```
fzyzcjy: nit: maybe also remove this since users just use `tms.region()`

torch_memory_saver/__init__.py
```diff
@@ -73,21 +109,19 @@ def compute():
...
 
-    def next_id(self):
-        self._last_id += 1
-        return self._last_id
-
 
 _global_info = _GlobalInfo()
 
+# Global singleton instance
+memory_saver = TorchMemorySaver()
```
fzyzcjy: nit: wondering whether or not we should name this variable `torch_memory_saver` (I am ok for both)

fzyzcjy (commented):
Good job! Some nits

csrc/torch_memory_saver.cpp
hebiao064: updated, good catch, thanks!

csrc/torch_memory_saver.cpp
hebiao064: updated, good catch, thanks!

torch_memory_saver/__init__.py
hebiao064: removed

torch_memory_saver/__init__.py
hebiao064: changed to `torch_memory_saver`

torch_memory_saver/__init__.py
hebiao064: removed

torch_memory_saver/__init__.py
hebiao064: yes, you are right, removed

torch_memory_saver/__init__.py
hebiao064: changed it back to `is_interesting_region`

csrc/torch_memory_saver.cpp
```diff
@@ -108,42 +109,72 @@ namespace CUDAUtils {
...
+        current_thread_context_.tag_ = tag;
+    }
+
+    bool is_interesting_region() const {
+        return current_thread_context_.is_interesting_region_;
+    }
+
+    void set_interesting_region(bool enabled) {
+        current_thread_context_.is_interesting_region_ = enabled;
+    }
```
fzyzcjy: I personally feel it good to make these in a separate class/namespace, for example originally I used a `namespace RegionManager` to separate the concerns

hebiao064: make sense, addressed, please review

csrc/torch_memory_saver.cpp
```diff
@@ -108,42 +109,83 @@ namespace CUDAUtils {
...
+}
+
 // ----------------------------------------------- primary class --------------------------------------------------
 
 struct _AllocationMetadata {
     size_t size;
     CUdevice device;
     CUmemGenericAllocationHandle allocHandle;
+    std::string tag;  // Add tag for memory type classification
```
fzyzcjy: tiny nit

```suggestion
    std::string tag;
```

csrc/torch_memory_saver.cpp
```diff
@@ -108,42 +109,83 @@ namespace CUDAUtils {
...
 
 class TorchMemorySaver {
 public:
-    TorchMemorySaver() {}
+    static TorchMemorySaver& getInstance() {
+        static TorchMemorySaver instance;
+        return instance;
+    }
+
+    // Delete copy constructor and assignment operator
```
fzyzcjy: tiny nit

```suggestion
```

csrc/torch_memory_saver.cpp
```diff
@@ -108,42 +109,83 @@ namespace CUDAUtils {
...
-
         CURESULT_CHECK(cuMemAddressReserve((CUdeviceptr *) ptr, size, 0, 0, 0));
         CURESULT_CHECK(cuMemMap((CUdeviceptr) * ptr, size, 0, allocHandle, 0));
         CUDAUtils::cu_mem_set_access(*ptr, size, device);
 
         {
-            const std::lock_guard <std::mutex> lock(allocator_metadata_mutex_);
-            allocation_metadata_.emplace(*ptr, _AllocationMetadata{size, device, allocHandle});
+            const std::lock_guard<std::mutex> lock(allocator_metadata_mutex_);
+            allocation_metadata_.emplace(*ptr, _AllocationMetadata{size, device, allocHandle, RegionManager::get_current_tag()});
```
fzyzcjy: nit: wondering whether we should make the tag be passed in as an arg. Because with this, the main TorchMemorySaver class is decoupled with the RegionManager (which is a global mutable state).

csrc/torch_memory_saver.cpp
```diff
@@ -210,60 +260,34 @@ class TorchMemorySaver {
...
 #endif
 
             metadata.allocHandle = newAllocHandle;
         }
     }
 
-    static TorchMemorySaver &instance() {
-        static TorchMemorySaver instance;
-        return instance;
-    }
```
fzyzcjy: nit: maybe keep this unchanged (not renamed + not moved) to make code diff minimized

fzyzcjy (commented):
LGTM, only a bit of nits!

csrc/torch_memory_saver.cpp
hebiao064: make sense, addressed, thanks

csrc/torch_memory_saver.cpp
hebiao064: make sense, addressed, thanks

csrc/torch_memory_saver.cpp
```diff
@@ -118,32 +119,32 @@ struct _AllocationMetadata {
...
     CUmemGenericAllocationHandle allocHandle;
+    std::string tag;
 };
 
 class TorchMemorySaver {
 public:
     TorchMemorySaver() {}
 
-    cudaError_t malloc(void **ptr, size_t size) {
+    cudaError_t malloc(void **ptr, size_t size, const std::string& tag = "default") {
```
fzyzcjy: nit: I personally like to make it non-default arg, b/c we always pass in something from upper level, but anyway this is just a nit

csrc/torch_memory_saver.cpp
```diff
@@ -166,39 +167,47 @@ class TorchMemorySaver {
...
-                  << " metadata.allocHandle=" << metadata.allocHandle
+                  << " metadata.allocHandle=" << metadata.allocHandle << " tag=" << metadata.tag
                   << std::endl;
 #endif
 
         return cudaSuccess;
     }
 
-    void pause() {
+    void pause(const std::string& tag = "") {
```
fzyzcjy: (same)

csrc/torch_memory_saver.cpp
```diff
@@ -166,39 +167,47 @@ class TorchMemorySaver {
...
                       << " ptr=" << ptr << " metadata.size=" << metadata.size << " metadata.allocHandle="
-                      << metadata.allocHandle
+                      << metadata.allocHandle << " tag=" << metadata.tag << " filter_tag=" << tag
                       << std::endl;
 #endif
         }
     }
 
-    void resume() {
+    void resume(const std::string& tag = "") {
```
fzyzcjy: (same)

torch_memory_saver/__init__.py
```diff
@@ -14,14 +14,14 @@
...
-        assert self._id == 1, 'Only support one single instance yet (multi-instance will be implemented later)'
 
     @contextmanager
-    def region(self):
+    def region(self, tag: str = "default"):
+        """Context manager for memory saving with optional tag"""
         if _global_info.binary_info.enabled:
             self._ensure_mem_pool()
             with torch.cuda.use_mem_pool(self._mem_pool):
+                _global_info.binary_info.cdll.tms_set_current_tag(tag.encode('utf-8'))
```
fzyzcjy, line 24: maybe reset tag back to default at the end of function, i.e. add `tms_set_currnet_tag(nullptr)` or whatever near `tms_region_leave`

fzyzcjy (commented):
Only some tiny nits (if you have time maybe spend 3 minute to change and I am ready to merge; if in a hurry I am also ok for current code)

fzyzcjy (approved):

fzyzcjy:
realize that, we may need one separate mem pool per tag. this is because, suppose:

* tag a and allocate 1KB tensor x: indeed torch allocate 2MB memory M, and slice it to satisfy x, so we tag M with a
* tag b and allocate 1KB tensor y: indeed torch reuse memory M, so tensor y is indeed with tag a

hebiao064:
> realize that, we may need one separate mem pool per tag in this mode. this is because, suppose:
> 
> * tag a and allocate 1KB tensor x: indeed torch allocate 2MB memory M, and slice it to satisfy x, so we tag M with a
> * tag b and allocate 1KB tensor y: indeed torch reuse memory M, so tensor y is indeed with tag a
> 
> I will also make a pluggable allocator version which may not have this issue though

Ah I think I do understand the mechanism (just read this blog few days ago: https://zhuanlan.zhihu.com/p/493646010)

But I wonder how did you realize this problem? and what kind of issues are you foreseeing right now? Will it lead to illegal memory issue we recently hit?

Context:
the illegal memory issue was mitigated by these two PRs
- https://github.com/sgl-project/sglang/pull/7733 
- https://github.com/sgl-project/sglang/pull/7812

fzyzcjy:
> But I wonder how did you realize this problem?

b/c today I am refactoring and adding features to torch_memory_saver, and this comes to my mind.

> and what kind of issues are you foreseeing right now?

I expect it to have issues like, tensors are wrongly tagged

hebiao064:
> > But I wonder how did you realize this problem?
> 
> b/c today I am refactoring and adding features to torch_memory_saver, and this comes to my mind.
> 
> > and what kind of issues are you foreseeing right now?
> 
> I expect it to have issues like, tensors are wrongly tagged

I'm happy to fix the pool arrangement issue, but tbh I am not very clear about the next.

Should I create separate pool when we init those tensors (e.g: kv and weight) and make sure we are malloc memory using the pool passed in?

PR #23, opened 2025-06-27
Add validations for allocation state transition and their calls.

MrAta:
Adding state management for memory allocations and a set of validation tests to ensure correct behavior under various scenarios by adding a `AllocationState` to allocation metadata.

### Validation Tests:
Run

```
python -m pytest test_examples.py::test_state_tracking_validation_basic -s
python -m pytest test_examples.py::test_state_tracking_validation_multi_tag -s
python -m pytest test_examples.py::test_state_tracking_validation_error_messages -s
```

### DIFF
--- a/csrc/core.cpp
+++ b/csrc/core.cpp
@@ -53,7 +53,7 @@ cudaError_t TorchMemorySaver::free(void *ptr) {
     return cudaSuccess;
 }
 
-void TorchMemorySaver::pause(const std::string& tag) {
+int TorchMemorySaver::pause(const std::string& tag) {
     const std::lock_guard <std::mutex> lock(allocator_metadata_mutex_);
 
     for (auto it = allocation_metadata_.begin(); it != allocation_metadata_.end(); ++it) {
@@ -65,9 +65,11 @@ void TorchMemorySaver::pause(const std::string& tag) {
         }
 
         if (metadata.state == AllocationState::PAUSED) {
-            std::string error_msg = "Cannot pause allocation that is already paused. "
-                                  "Tag: " + metadata.tag + ", Ptr: " + std::to_string((uintptr_t)ptr);
-            throw std::runtime_error(error_msg);
+            std::cerr << "[torch_memory_saver.cpp] Cannot pause allocation that is already paused. "
+                      << "Tag: " << metadata.tag << ", Ptr: " << std::to_string((uintptr_t)ptr)
+                      << " file=" << __FILE__ << " func=" << __func__ << " line=" << __LINE__
+                      << std::endl;
+            return 1;
         }
 
         if (metadata.enable_cpu_backup) {
@@ -82,7 +84,6 @@ void TorchMemorySaver::pause(const std::string& tag) {
         CURESULT_CHECK(cuMemUnmap((CUdeviceptr) ptr, metadata.size));
         CURESULT_CHECK(cuMemRelease(metadata.allocHandle));
 
-        // Update state to paused
         metadata.state = AllocationState::PAUSED;
 
 #ifdef TMS_DEBUG_LOG
@@ -93,9 +94,11 @@ void TorchMemorySaver::pause(const std::string& tag) {
                   << std::endl;
 #endif
     }
+
+    return 0;
 }
 
-void TorchMemorySaver::resume(const std::string& tag) {
+int TorchMemorySaver::resume(const std::string& tag) {
     const std::lock_guard <std::mutex> lock(allocator_metadata_mutex_);
 
     for (auto it = allocation_metadata_.begin(); it != allocation_metadata_.end(); ++it) {
@@ -107,9 +110,11 @@ void TorchMemorySaver::resume(const std::string& tag) {
         }
 
         if (metadata.state == AllocationState::ACTIVE) {
-            std::string error_msg = "Cannot resume allocation that is already active. "
-                                  "Tag: " + metadata.tag + ", Ptr: " + std::to_string((uintptr_t)ptr);
-            throw std::runtime_error(error_msg);
+            std::cerr << "[torch_memory_saver.cpp] Cannot resume allocation that is already active. "
+                      << "Tag: " << metadata.tag << ", Ptr: " << std::to_string((uintptr_t)ptr)
+                      << " file=" << __FILE__ << " func=" << __func__ << " line=" << __LINE__
+                      << std::endl;
+            return 1;
         }
 
         CUmemGenericAllocationHandle newAllocHandle;
@@ -126,9 +131,6 @@ void TorchMemorySaver::resume(const std::string& tag) {
             // maybe we can free host memory if needed (currently keep it there to reduce re-alloc time)
         }
 
-        metadata.state = AllocationState::ACTIVE;
-        metadata.allocHandle = newAllocHandle;
-
 #ifdef TMS_DEBUG_LOG
         std::cout << "[torch_memory_saver.cpp] TorchMemorySaver.resume"
                   << " ptr=" << ptr << " metadata.size=" << metadata.size << " (old)metadata.allocHandle="
@@ -137,5 +139,9 @@ void TorchMemorySaver::resume(const std::string& tag) {
                   << " metadata.enable_cpu_backup=" << metadata.enable_cpu_backup
                   << std::endl;
 #endif
+        metadata.state = AllocationState::ACTIVE;
+        metadata.allocHandle = newAllocHandle;
     }
+
+    return 0;
 }
--- a/csrc/core.cpp
+++ b/csrc/core.cpp
@@ -64,14 +64,10 @@ void TorchMemorySaver::pause(const std::string& tag) {
             continue;
         }
 
-        // Check if already paused
         if (metadata.state == AllocationState::PAUSED) {
-#ifdef TMS_DEBUG_LOG
-            std::cout << "[torch_memory_saver.cpp] TorchMemorySaver.pause SKIPPED (already paused)"
-                      << " ptr=" << ptr << " tag=" << metadata.tag << " filter_tag=" << tag
-                      << std::endl;
-#endif
-            continue;  // Skip if already paused
+            std::string error_msg = "Cannot pause allocation that is already paused. "
+                                  "Tag: " + metadata.tag + ", Ptr: " + std::to_string((uintptr_t)ptr);
+            throw std::runtime_error(error_msg);
         }
 
         if (metadata.enable_cpu_backup) {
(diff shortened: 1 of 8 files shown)


csrc/torch_memory_saver.cpp
```diff
@@ -166,27 +180,41 @@ class TorchMemorySaver {
...
+                  << " ptr=" << ptr << " metadata.size=" << metadata.size << " metadata.allocHandle="
+                  << metadata.allocHandle << " tag=" << metadata.tag
                   << std::endl;
 #endif
 
         return cudaSuccess;
     }
 
-    void pause(const std::string& tag) {
+    int pause(const std::string& tag) {
```
fzyzcjy: nit: maybe we just directly print+exit(1) on failures to simplify code, since users should never do that (unless the code is buggy) and it seems OK to exit on that case

fzyzcjy (commented):
Looks good! Only a nit (on this round of review)

csrc/torch_memory_saver.cpp
```diff
@@ -195,11 +223,24 @@ class TorchMemorySaver {
                       << std::endl;
 #endif
         }
+
+        if (!found_valid_allocation) {
```
hebiao064: is it possible to reach this scenario? if there is no pausable allocation, we can still return 0?

csrc/torch_memory_saver.cpp
```diff
@@ -208,23 +249,39 @@ class TorchMemorySaver {
...
+
+        if (!found_paused_allocation) {
+            std::string error_msg = "No paused allocations found";
+            if (!tag.empty()) {
+                error_msg += " for tag: " + tag;
+            }
+            error_msg += ". Cannot resume unpaused or non-existent allocations.";
+            std::cerr << "[torch_memory_saver.cpp] ERROR: " << error_msg << std::endl;
+            return -1;
+        }
```
hebiao064: same, i think it might not be necessary

examples/test_state_validation.py
```diff
@@ -0,0 +1,162 @@
...
+    except RuntimeError as e:
+        if "No active allocations found" in str(e):
+            print("SUCCESS: Pause after free correctly succeeded (no allocations to pause)")
+            return True
+        else:
+            print(f"ERROR: Unexpected exception: {e}")
+            return False
+    except Exception as e:
+        print(f"ERROR: Unexpected exception: {e}")
+        return False
```
hebiao064: how about

```
    try:
        torch_memory_saver.pause()
    except RuntimeError as e:
        assert "No active allocations found" in str(e), f"Unexpected RuntimeError: {e}"

print("SUCCESS: Pause after tensor freed correctly failed with expected error msg (no allocations to pause)")
```

torch_memory_saver/__init__.py
```diff
@@ -35,13 +35,23 @@ def pause(self, tag: Optional[str] = None):
         """Pause memory for specific tag or all memory if tag is None"""
         if _global_info.binary_info.enabled:
             tag_bytes = tag.encode('utf-8') if tag else None
-            _global_info.binary_info.cdll.tms_pause(tag_bytes)
+            result = _global_info.binary_info.cdll.tms_pause(tag_bytes)
```
hebiao064: if the error msg was logged by cpp code, maybe we don't need to log it again from the python code?

csrc/torch_memory_saver.cpp
MrAta: I tried that approach, but, he problem is that if we exit(1) from the .so the whole process exists and the call from python cannot catch it. So for better error handling and clear message, we need to have a return type here.

csrc/torch_memory_saver.cpp
MrAta: Yes, if we call pause on a region that has no active allocation then it indicates a silent misuse of the API calls. So I think it'd be nice to catch that for better debugging experience for the down stream users.

csrc/torch_memory_saver.cpp
MrAta: see my previous comment.

torch_memory_saver/__init__.py
MrAta: See my reply to Tom's comment above.

csrc/torch_memory_saver.cpp
fzyzcjy: yes, but I do not expect python to catch it - this is a bug and programmers should fix it in their code and rerun the program and see no error, thus I slightly prefer making the code simpler

csrc/torch_memory_saver.cpp
```diff
@@ -208,22 +248,36 @@ class TorchMemorySaver {
                 continue;
             }
 
-            CUmemGenericAllocationHandle newAllocHandle;
-            CUDAUtils::cu_mem_create(&newAllocHandle, metadata.size, metadata.device);
+            if (metadata.state == AllocationState::PAUSED) {
```
fzyzcjy: nit: for symmetry, it seems that when we pause a paused memory we error, then what about we also error when we resume a resumed memory

torch_memory_saver/__init__.py
```diff
@@ -39,13 +39,19 @@ def pause(self, tag: Optional[str] = None):
         """Pause memory for specific tag or all memory if tag is None"""
         if _global_info.binary_info.enabled:
             tag_bytes = tag.encode("utf-8") if tag else None
-            _global_info.binary_info.cdll.tms_pause(tag_bytes)
+            try:
+                _global_info.binary_info.cdll.tms_pause(tag_bytes)
+            except Exception as e:
+                raise e
```
fzyzcjy: nit: maybe we can remove the try-except

torch_memory_saver/__init__.py
```diff
@@ -39,13 +39,19 @@ def pause(self, tag: Optional[str] = None):
...
 
     def resume(self, tag: Optional[str] = None):
         """Resume memory for specific tag or all memory if tag is None"""
         if _global_info.binary_info.enabled:
             tag_bytes = tag.encode("utf-8") if tag else None
-            _global_info.binary_info.cdll.tms_resume(tag_bytes)
+            try:
+                _global_info.binary_info.cdll.tms_resume(tag_bytes)
+            except Exception as e:
+                raise e
```
fzyzcjy: (same)

csrc/torch_memory_saver.cpp
```diff
@@ -208,22 +248,36 @@ class TorchMemorySaver {
...
+            }
+        }
+
+        if (!found_paused_allocation) {
+            std::string error_msg = "No paused allocations found";
+            if (!tag.empty()) {
+                error_msg += " for tag: " + tag;
+            }
+            error_msg += ". Cannot resume unpaused or non-existent allocations.";
+            std::cerr << "[torch_memory_saver.cpp] ERROR: " << error_msg << std::endl;
```
fzyzcjy: nit: what about one line

```
std::cerr << .... << tag << ...
```

no worries, even if tag is empty, we can still print that tag

csrc/torch_memory_saver.cpp
```diff
@@ -195,11 +223,23 @@ class TorchMemorySaver {
...
 #endif
         }
+
+        if (!found_valid_allocation) {
+            std::string error_msg = "No active allocations found";
+            if (!tag.empty()) {
+                error_msg += " for tag: " + tag;
+            }
+            error_msg += ". Cannot pause paused or non-existent allocations.";
+            std::cerr << "[torch_memory_saver.cpp] ERROR: " << error_msg << std::endl;
```
fzyzcjy: (same)

csrc/torch_memory_saver.cpp
```diff
@@ -177,16 +191,30 @@ class TorchMemorySaver {
...
                 continue;
             }
 
+            if (metadata.state == AllocationState::PAUSED) {
+                std::string error_msg = "Trying to pause already paused allocation";
+                if (!tag.empty()) {
+                    error_msg += " for tag: " + tag;
+                }
+                error_msg += " ptr=" + std::to_string(reinterpret_cast<uintptr_t>(ptr));
+                std::cerr << "[torch_memory_saver.cpp] ERROR: " << error_msg << std::endl;
```
fzyzcjy: similar about this (and other things), maybe just a `std::cerr << ...` and no std::string msg

fzyzcjy (commented):
LGTM! Only some nits to simplify the code a bit

csrc/core.cpp
```diff
@@ -111,6 +126,9 @@ void TorchMemorySaver::resume(const std::string& tag) {
             // maybe we can free host memory if needed (currently keep it there to reduce re-alloc time)
         }
 
+        metadata.state = AllocationState::ACTIVE;
+        metadata.allocHandle = newAllocHandle;
```
fzyzcjy: if this line is moved to here, then our " (old)metadata.allocHandle=" log should be changed, o/w we always print new handle

csrc/core.cpp
```diff
@@ -76,6 +82,9 @@ void TorchMemorySaver::pause(const std::string& tag) {
         CURESULT_CHECK(cuMemUnmap((CUdeviceptr) ptr, metadata.size));
         CURESULT_CHECK(cuMemRelease(metadata.allocHandle));
 
+        // Update state to paused
```
fzyzcjy: nit: no need for comments for this

csrc/core.cpp
```diff
@@ -64,6 +64,12 @@ void TorchMemorySaver::pause(const std::string& tag) {
             continue;
         }
 
+        if (metadata.state == AllocationState::PAUSED) {
+            std::string error_msg = "Cannot pause allocation that is already paused. "
+                                  "Tag: " + metadata.tag + ", Ptr: " + std::to_string((uintptr_t)ptr);
+            throw std::runtime_error(error_msg);
```
fzyzcjy: hmm, I am worried sometimes c++ exceptions are disabled, e.g. google c++ guide https://google.github.io/styleguide/cppguide.html#Exceptions

and about https://github.com/fzyzcjy/torch_memory_saver/pull/23#discussion_r2190703876

> @MrAta : I tried that approach, but, he problem is that if we exit(1) from the .so the whole process exists and the call from python cannot catch it. So for better error handling and clear message, we need to have a return type here.

It seems I often see assertions on c++ code that directly make program halt. For example, DeepEP, nvshmem, etc. So maybe it is not a big problem, unless users want to recover from the error (but it seems not that case).

For testability, yes I totally agree about that, but for this minimal case maybe we can just reuse the current test infra that each test run in a subprocess and we may check the exit code of the subprocess.

csrc/entrypoint.cpp
```diff
@@ -11,6 +13,9 @@ struct ThreadLocalConfig {
 };
 static thread_local ThreadLocalConfig thread_local_config;
 
+static std::string last_error_message;
```
fzyzcjy: (same for all these - would be great if we skip error handling to keep code minimal)

csrc/core.cpp
MrAta: Good catch! Fixed it!

csrc/core.cpp
MrAta: Yes, agreed! I updated the code to use an int return type for the pause and resume instead of exiting altogether; Please take a look and let me know if that looks good to you. If not, I can fall back to exit(1) method.

fzyzcjy: To be honest I hope it is just exit(1), since this seems to be a bug and users should directly fix it and it is ok to directly halt, instead of some recoverable exception. Or if it is exit code, maybe need to use it for every call to be consistent (which is verbose).

PR #45, opened 2025-08-12
release `cpu_backup` to avoid memory leak on cpu

constroy:


constroy:
> Great catch! btw curious where do you use torch_memory_saver?

Just for study.
I am interested in memory allocators and PyTorch internals.

fzyzcjy:
I see :)

PR #48, opened 2025-08-26
[BugFix] support GDR if possible

reyoung:
The current `master` version of `torch_memory_saver` does not support GPUDirect RDMA transport.
`ibv_reg_mr` will fail and report `bad address` when using memory saver.

This PR fix this bug.

### DIFF
--- a/csrc/utils.h
+++ b/csrc/utils.h
@@ -184,6 +184,13 @@ namespace CUDAUtils {
         prop.type = CU_MEM_ALLOCATION_TYPE_PINNED;
         prop.location.type = CU_MEM_LOCATION_TYPE_DEVICE;
         prop.location.id = device;
+
+        int flag = 0;
+        CURESULT_CHECK(cuDeviceGetAttribute(&flag, CU_DEVICE_ATTRIBUTE_GPU_DIRECT_RDMA_WITH_CUDA_VMM_SUPPORTED, device));
+        if (flag) {  // support GPUDirect RDMA if possible
+            prop.allocFlags.gpuDirectRDMACapable = 1;
+        }
+
         CURESULT_CHECK(cuMemCreate(alloc_handle, size, &prop, 0));
     }
 


reyoung:
> Looks reasonable. curious have you tested whether it supports release & resume? IIRC rdma requires unchanged physical memory, and then this approach will fail.

RDMA-like api tested by using `nixl`.

Just un-reg memory regions before release, and call reg_mr after resume.

Sample code like

---

server.py

```python
import time

import torch
from nixl._api import nixl_agent, nixl_agent_config
from torch_memory_saver import torch_memory_saver


def main():
    torch_memory_saver.hook_mode = "preload"
    buffer1 = torch.ones((1024, 1024, 4), dtype=torch.uint8, device="cuda:0")
    agent = nixl_agent("svr", nixl_agent_config())
    desc1 = agent.register_memory(buffer1)

    agent.deregister_memory(desc1)
    print("unregistered buffer1")
    torch_memory_saver.pause()
    torch_memory_saver.resume()
    buffer1.fill_(7)
    torch.cuda.synchronize()
    desc1 = agent.register_memory(buffer1)

    with open("agent.meta", "wb") as f:
        f.write(agent.get_agent_metadata())
    with open("param.meta", "wb") as f:
        f.write(agent.get_serialized_descs(desc1.trim()))
    time.sleep(3600)

    agent.deregister_memory(desc1)


if __name__ == "__main__":
    main()
```

---
client.py

```
import time

import torch
from nixl._api import nixl_agent, nixl_agent_config
from torch_memory_saver import torch_memory_saver


def main():
    torch_memory_saver.hook_mode = "preload"
    agent = nixl_agent("cli", nixl_agent_config())
    buf = torch.zeros((1024, 1024, 4), dtype=torch.uint8, device="cuda:1")

    ldesc = agent.register_memory(buf)
    agent.deregister_memory(ldesc)
    torch_memory_saver.pause()
    torch_memory_saver.resume()
    ldesc = agent.register_memory(buf)

    with open("agent.meta", "rb") as f:
        remote_name = agent.add_remote_agent(f.read())
    with open("param.meta", "rb") as f:
        rdesc = agent.deserialize_descs(f.read())

    xfer = agent.initialize_xfer("READ", ldesc.trim(), rdesc, remote_name)
    print(agent.check_remote_metadata(remote_name, rdesc))
    agent.transfer(xfer)

    while agent.check_xfer_state(xfer) == "PROC":
        time.sleep(0.5)
    print(buf)
    print(agent.check_xfer_state(xfer))


if __name__ == "__main__":
    main()
```

reyoung:
BTW, I think `nixl` might be fit for RL parameter distribution for the following reasons

1. `nixl` does not require creating `ncclComm` and avoids the cost of `ncclComm`.
2. The API is straightforward.
    1. Single-Side `Read/Write` operation to avoid send/recv on both sides.
    2. Operations also contain a `notification` to notify the network that the operation is done. RL trainer can offload parameters as soon as possible when the rollout read operation is done.
4. Scatter/Gather is supported natively by `nixl`. No gather/scatter are required.

fzyzcjy:
> Just re-register memory regions 

Ok then there is no problem, it should work well! (I was talking about not re-registering)

PR #50, opened 2025-10-05
fix cudart deps

leng-yue:
This PR fixed the following error:

```
torch_memory_saver_hook_mode_preload.abi3.so: undefined symbol: cudaMallocHost
```


fzyzcjy:
Great job! btw I am curious why it did work on my machine (and many others' machine) without this line...

PR #52, opened 2025-10-08
[Hardware Support] AMD - version update: ROCm-6.3.4~7.0; GPUs-6.3.4~7.0

yushengsu-thu:
[Done]
- Fixed the compile issue in setup.py for version 0.0.9rc2.
- Added support for ROCm versions: 6.3.4 ~ 7.0.0.
- Added support for AMD GPU series: MI300~MI325
- Added support for CPU and pass test: `cpu_backup.py` as well.
- Because the current hipmem API (analogous to CUDA's cuMem) still has issues, I implemented a chunk-wise approach and isolated the implementation details in separate files: `hardware_amd_support.cpp` and `hardware_amd_support.h`.

[Test]
Screenshot 2025-10-08 at 4 21 25 PM

[To-do]
- Continue coordinating with the AMD ROCm team to fully enable the hipmem API and align the implementation with the CUDA side.
- Add support for MI350.


csrc/core.cpp
```diff
@@ -309,9 +220,7 @@ void TorchMemorySaver::resume(const std::string& tag) {
 
         if (metadata.enable_cpu_backup) {
             SIMPLE_CHECK(metadata.cpu_backup != nullptr, "cpu_backup should not be nullptr");
-            // TODO may use cudaMemcpyAsync if needed
             CUDA_ERROR_CHECK(cudaMemcpy(ptr, metadata.cpu_backup, metadata.size, cudaMemcpyHostToDevice));
-            // maybe we can free host memory if needed (currently keep it there to reduce re-alloc time)
```
fzyzcjy, line 314: hmm maybe we can only change #if AMD part and make cuda part char-by-char unchanged

fzyzcjy:
LGTM as long as no CUDA or common code is changed, i.e. only AMD code is changed

csrc/core.cpp
yushengsu-thu: Resolved

PR #55, opened 2025-10-30
Fix wrong mode attached to buffers

fzyzcjy:
fix this -> https://github.com/fzyzcjy/torch_memory_saver/pull/20#issuecomment-3047099047


fzyzcjy:
confirmed by @zhuzilin can run

PR #62, opened 2025-11-27
cudaFree should always go through the tms API.

66RING:
Tensor may be free outside of the context manager. And the vmem will be free by original cudaFree for LD_PRELOAD mode.

```cpp
cudaError_t cudaFree(void *ptr) {
    if (thread_local_config.is_interesting_region()) {
        return TorchMemorySaver::instance().free(ptr);
    } else {
        // cudaFree a vmem ptr!
        return APIForwarder::call_real_cuda_free(ptr);
    }
}
```
Some thing like:

```python
with torch_memory_saver.region():
    pauseable_tensor = torch.full((1_000_000_000,), 100, dtype=torch.uint8, device='cuda')

# pauseable_tensor been free some whare. but not in any interesting region.
del pauseable_tensor
```

### DIFF
--- a/csrc/core.cpp
+++ b/csrc/core.cpp
@@ -212,3 +212,35 @@ void TorchMemorySaver::resume(const std::string& tag) {
     #error "USE_PLATFORM is not set"
 #endif
 }
+
+uint8_t* TorchMemorySaver::get_cpu_backup_pointer(const uint8_t* query_gpu_ptr, uint64_t query_size) {
+#if defined(USE_ROCM)
+    exit(1); // unsupported
+
+#elif defined(USE_CUDA)
+    const std::lock_guard <std::mutex> lock(allocator_metadata_mutex_);
+
+    for (auto it = allocation_metadata_.begin(); it != allocation_metadata_.end(); ++it) {
+        uint8_t *ptr = (uint8_t*) it->first;
+        AllocationMetadata &metadata = it->second;
+
+        if ((ptr <= query_gpu_ptr) && (query_gpu_ptr + query_size <= ptr + metadata.size)) {
+            const size_t offset = query_gpu_ptr - ptr;
+            if (metadata.state == AllocationState::ACTIVE) {
+                return nullptr;
+            } else {
+                SIMPLE_CHECK(nullptr != metadata.cpu_backup,
+                    "get_cpu_backup_pointer: found paused allocation but cpu_backup does not exist, do you forget to enable cpu backup");
+                return (uint8_t*) metadata.cpu_backup + offset;
+            }
+        }
+    }
+
+    std::cerr << "[torch_memory_saver.cpp] get_cpu_backup_pointer fail to find backup "
+              << " query_gpu_ptr=" << query_gpu_ptr << " query_size=" << query_size
+              << std::endl;
+    exit(1);
+#else
+    #error "USE_PLATFORM is not set"
+#endif
+}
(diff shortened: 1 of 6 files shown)


fzyzcjy:
Good job, could you please add a few lines in the tests to cover this case

66RING:
> Good job, could you please add a few lines in the tests to cover this case

@fzyzcjy Maybe just check exitcode in test where all exitcode is expected to be 0.

But torch allocator looks buggy https://github.com/pytorch/pytorch/issues/145168 .since "gc" will be called at the end of python proccess. And the "gc" may call cudaFree outside of MemPool. Thus, "gc" free a vptr using cudaFree and then segment fault at `c10::cuda::CUDACachingAllocator::Native::DeviceCachingAllocator::release_block`.

Before this pr:

```bash
[torch_memory_saver]% pytest test/test_examples.py::test_simple                                                                                 
=========================================================================== test session starts ============================================================================
platform linux -- Python 3.13.7, pytest-9.0.1, pluggy-1.6.0
rootdir: /home/ring/t/torch_memory_saver
collected 2 items                                                                                                                                                          

test/test_examples.py F.                                                                                                                                             [100%]
```
Cuda error is reported but still `output_queue.put(True)` without raise Exception

```
terminate called after throwing an instance of 'c10::AcceleratorError'
  what():  CUDA error: invalid argument
Search for `cudaErrorInvalidValue' in https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html for more information.
CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
For debugging consider passing CUDA_LAUNCH_BLOCKING=1
Compile with `TORCH_USE_CUDA_DSA` to enable device-side assertions.

Exception raised from c10_cuda_check_implementation at /pytorch/c10/cuda/CUDAExceptio
frame #0: c10::Error::Error(c10::SourceLocation, std::__cxx11::basic_string<char, std
nvironments/tms/lib/python3.13/site-packages/torch/lib/libc10.so)
frame #1: <unknown function> + 0x11fb7 ............
frame #2: <unknown function> + 0x202d8 ............
frame #3: <unknown function> + 0x3a0c9 ............
frame #4: c10::cuda::MemPool::~MemPool() + 0x1bc ............
frame #5: <unknown function> + 0xd4a112 ............
frame #6: <unknown function> + 0x3c192c ............
frame #7: <unknown function> + 0x3c1fbe ............
<omitting python frames>
frame #22: <unknown function> + 0x27635 (0x797cc9427635 in /usr/lib/libc.so.6)
frame #23: __libc_start_main + 0x89 (0x797cc94276e9 in /usr/lib/libc.so.6)
frame #24: _start + 0x25 ............

```

After this pr:
```bash
[torch_memory_saver]% pytest test/test_examples.py::test_simple     
=========================================================================== test session starts ============================================================================
platform linux -- Python 3.13.7, pytest-9.0.1, pluggy-1.6.0
rootdir: /home/ring/t/torch_memory_saver
collected 2 items                                                                                                                                                          

test/test_examples.py ..                                                                                                                                             [100%]

============================================================================ 2 passed in 11.57s ============================================================================
```

fzyzcjy:
Looks reasonable. Just a final check - do you run all tests i.e. `pytest test/test_examples.py` and they pass? (just to ensure no regression since we do not have ci)

66RING:
> Looks reasonable. Just a final check - do you run all tests i.e. `pytest test/test_examples.py` and they pass? (just to ensure no regression since we do not have ci)

Yes, I had test all ut `pytest test/` and all pass.

PR #64, opened 2025-11-28
Fix: Handle multiple libraries in LD_PRELOAD for torch_memory_saver

XcodeRole:
Handle multiple libraries in LD_PRELOAD

### DIFF
--- a/torch_memory_saver/hooks/mode_preload.py
+++ b/torch_memory_saver/hooks/mode_preload.py
@@ -11,20 +11,14 @@ class HookUtilModePreload(HookUtilBase):
     def get_path_binary(self):
         env_ld_preload = os.environ.get("LD_PRELOAD", "")
         
-        ld_preload_paths = env_ld_preload.split(":")
-        torch_memory_saver_path = None
-        for path in ld_preload_paths:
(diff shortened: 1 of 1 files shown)


torch_memory_saver/hooks/mode_preload.py
```diff
@@ -10,13 +10,21 @@
...
     def get_path_binary(self):
         env_ld_preload = os.environ.get("LD_PRELOAD", "")
-        assert "torch_memory_saver" in env_ld_preload, (
+        
+        ld_preload_paths = env_ld_preload.split(":")
+        torch_memory_saver_path = None
+        for path in ld_preload_paths:
+            if "torch_memory_saver" in path:
+                torch_memory_saver_path = path
+                break
```
fzyzcjy: nit: what about

```
interest_paths = [p for p in env_ld_preload.split(":") if "torch_memory_saver" in path]
assert len(interest_paths) == 1, "the error message like before"
return interest_paths[0]
```

XcodeRole: Thanks for the good suggestion! Much cleaner — will adopt this.

PR #74, opened 2026-04-20
Ship cu12 and cu13 .so variants in one wheel (0.0.9.post1)

Kangyan-Zhou:
## Motivation

The PyPI `torch_memory_saver==0.0.9` wheel is linked only against `libcudart.so.12`. In a CUDA 13 runtime, `LD_PRELOAD` of the `.so` fails at import:

\`\`\`
python3: error while loading shared libraries: libcudart.so.12: cannot open shared object file
\`\`\`

This breaks every `enable_memory_saver=True` path under CUDA 13 in downstream projects — notably sglang, whose `dev-cu13` CI pipeline currently has to paper over it (tests were parked in `test/manual/` with TODO banners pointing back here).

## What this PR does

Rebuilds the same 0.0.9 source tree to emit **four** `.so` files per arch — `torch_memory_saver_hook_mode_{preload,torch}_cu{12,13}.abi3.so` — and selects the correct one at import time via `torch.version.cuda` (with a `ctypes.CDLL("libcudart.so.{13,12}")` probe fallback). A single wheel now covers both CUDA 12 and 13 runtimes with no downstream install-time gymnastics.

Released as **`0.0.9.post1`** (PEP 440 post-release = source-equivalent minor correction) rather than `0.0.10`, since the Python API and C++ sources are unchanged — only the build pipeline and runtime resolver are new. Downstream projects pinning `==0.0.9` do not pick up the change by default; consumers that need CUDA 13 support opt in via `>=0.0.9.post1`.

### Runtime

- **`setup.py`** — version bump to `0.0.9.post1`. Extension names suffixed with `_cu{major}` via `TMS_CUDA_MAJOR` env var, required for CUDA builds (set by `scripts/build_in_docker.sh`). ROCm stays single-variant, unchanged.
- **`torch_memory_saver/utils.py`** — new `_detect_cuda_major()` (torch.version.cuda → libcudart probe) and `get_binary_path_from_package` now picks `{stem}_cu{major}.*.so`.

### Build pipeline

- **`scripts/build_multi_cuda.sh`** (new) — orchestrates two `scripts/build.sh` runs (CUDA 12.8 + 13.0) and merges the resulting per-CUDA wheels.
- **`scripts/merge_cuda_wheels.py`** (new) — zips the `.so` contents of two input wheels into one, unions `top_level.txt`, regenerates RECORD, drops the `+cuXXX` local-version tag from the filename.
- **`scripts/build_in_docker.sh`** — drops the dead `pip install torch==...` step (tms's C++ sources don't use torch or nvcc) and sets `TMS_CUDA_MAJOR` from `CUDA_VERSION`. Simpler, faster builds; also sidesteps per-CUDA-version torch availability mismatches.
- **`scripts/build.sh`** — `ARCH` respects caller env (`${ARCH:-$(uname -m)}`) so the same script runs on BSD/macOS build hosts where `uname -i` isn't supported (e.g. Apple Silicon for aarch64 wheels).
- **`scripts/rename_wheels.sh`** — skip already-renamed wheels (prevented double-suffix output when multiple per-CUDA builds accumulate in `dist/`) and drive the per-CUDA tag from `CUDA_VERSION` env var rather than `ls /usr/local/ | grep`.
- **`Makefile`** — new `build-wheel-multi-cuda[-aarch64]` targets.

### Single-variant legacy path

`TMS_CUDA_MAJOR` is required for CUDA builds. Plain `pip install .` from an sdist fails with a clear error; the official distribution path for this release is PyPI wheels built via `scripts/build_multi_cuda.sh`. (sdist could be re-enabled later with a sensible `TMS_CUDA_MAJOR` default if users want source installs; not scoped here.)

## Verification

Built on:

- **x86_64** — `pytorch/manylinux2_28-builder:cuda12.8` + `:cuda13.0` on a Linux H200 host.
- **aarch64** — `pytorch/manylinuxaarch64-builder:cuda12.8` + `:cuda13.0` natively on Apple Silicon (Docker Desktop arm64 VM).

Both wheels carry all 8 expected `.so` files with correct NEEDED entries:

| Arch | Hook | CUDA | NEEDED |
|---|---|---|---|
| x86_64 | preload | cu12 | libcudart.so.12 ✅ |
| x86_64 | preload | cu13 | libcudart.so.13 ✅ |
| x86_64 | torch   | cu12 | libcudart.so.12 ✅ |
| x86_64 | torch   | cu13 | libcudart.so.13 ✅ |
| aarch64 | preload | cu12 | libcudart.so.12 ✅ |
| aarch64 | preload | cu13 | libcudart.so.13 ✅ |
| aarch64 | torch   | cu12 | libcudart.so.12 ✅ |
| aarch64 | torch   | cu13 | libcudart.so.13 ✅ |

**End-to-end on H200 + CUDA 13.0.1 + `lmsysorg/sglang:dev-cu13`:**

- Runtime detection resolves `get_binary_path_from_package("torch_memory_saver_hook_mode_preload")` and `..._torch` to the `_cu13.abi3.so` variants.
- `sglang test_utils_update_weights` (Llama-3.2-1B, `enable_memory_saver=True`, 1 GPU) **PASSED** in 45s.
- `sglang test_multi_instance_release_memory_occupation` (dp=2 tp=2, Qwen-1.5B, full `release_memory_occupation` → HF reload → `update_weights_from_tensor` → `resume_memory_occupation` across 4 ranks) **PASSED** in 71s.

## Artifacts

Draft / pre-upload wheels available at sha256:

- `torch_memory_saver-0.0.9.post1-cp39-abi3-manylinux2014_x86_64.whl` — `35c7266d265663f10985f99db2dfebdbe53ca97f5a8f134a2ac29e08066d31b3`
- `torch_memory_saver-0.0.9.post1-cp39-abi3-manylinux2014_aarch64.whl` — `60c087d603ed5e4d5b653fe7849417607b0aaea38f2d096bac444555b79928e0`

Happy to share via any channel you prefer (direct handoff, attached to a draft release on my fork, etc.) so you can verify locally and upload to PyPI from your account.

## Follow-up

A companion PR on sgl-project/sglang [#23182](https://github.com/sgl-project/sglang/pull/23182) bumps the pin to `torch_memory_saver>=0.0.9.post1` and re-registers the two tests that were parked in `test/manual/` because of this bug. That PR is also draft, gated on this release being published.

### DIFF
--- a/torch_memory_saver/utils.py
+++ b/torch_memory_saver/utils.py
@@ -1,3 +1,4 @@
+import ctypes
 import logging
 import os
 from contextlib import contextmanager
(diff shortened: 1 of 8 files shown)


torch_memory_saver/utils.py
```diff
@@ -1,15 +1,51 @@
...
 import os
 from contextlib import contextmanager
 from pathlib import Path
 
 logger = logging.getLogger(__name__)
 
+_SUPPORTED_CUDA_MAJORS = (13, 12)
+
+
+def _detect_cuda_major() -> int:
```
fzyzcjy, line 12: nit: would be great to have UT

PR #76, opened 2026-05-08
Fix ROCm binary resolution

TideDra:
## Summary
- Detect ROCm/HIP torch builds before probing CUDA runtimes.
- Resolve unsuffixed ROCm torch_memory_saver shared libraries while preserving CUDA _cuXX lookup.
- Add resolver tests covering ROCm and CUDA paths.

## Tests
- python -m pytest test/test_utils.py test/test_configure_subprocess.py -q

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -113,7 +113,7 @@ def _create_ext_modules(platform):
     # Suffix the compiled .so files with the CUDA major they link against, so
     # a single wheel can ship cu12 and cu13 variants side-by-side. `utils.py`
     # picks the right one at runtime via torch.version.cuda / libcudart probe.
-    # ROCm builds are single-variant for now (no suffix).
+    # ROCm builds are single-variant for now and use unsuffixed binaries.
     if platform == "cuda":
         cuda_major = os.environ.get("TMS_CUDA_MAJOR")
         if not cuda_major:
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -1,0 +1,34 @@
+import sys
+from types import SimpleNamespace
+
+from torch_memory_saver import utils
+
+
+def _fake_torch(cuda=None, hip=None):
+    return SimpleNamespace(version=SimpleNamespace(cuda=cuda, hip=hip))
+
+
+def test_get_binary_path_from_package_uses_unsuffixed_rocm_binary(monkeypatch, tmp_path):
+    stem = "torch_memory_saver_hook_mode_preload"
+    package_dir = tmp_path / "torch_memory_saver"
+    package_dir.mkdir()
+    binary = tmp_path / f"{stem}.abi3.so"
+    binary.touch()
+
+    monkeypatch.setattr(utils, "__file__", str(package_dir / "utils.py"))
+    monkeypatch.setitem(sys.modules, "torch", _fake_torch(hip="7.2.0"))
+
+    assert utils.get_binary_path_from_package(stem) == binary
+
+
+def test_get_binary_path_from_package_keeps_cuda_suffix_selection(monkeypatch, tmp_path):
+    stem = "torch_memory_saver_hook_mode_preload"
+    package_dir = tmp_path / "torch_memory_saver"
+    package_dir.mkdir()
+    binary = tmp_path / f"{stem}_cu12.abi3.so"
+    binary.touch()
+
+    monkeypatch.setattr(utils, "__file__", str(package_dir / "utils.py"))
+    monkeypatch.setitem(sys.modules, "torch", _fake_torch(cuda="12.9"))
+
+    assert utils.get_binary_path_from_package(stem) == binary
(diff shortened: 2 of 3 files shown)


fzyzcjy:
quick question: may I know what error do you see before fixing it? iirc the lib is already used by AMD cases (e.g. by @yushengsu-thu)

TideDra:
I encounter the following error when running slime:
```

 (SGLangEngine pid=244598)   File "/workspace/gongrui/.uv_python_cache/cpython-3.12.13-linux-x86_64-gnu/lib/python3.12/multiprocessing/process.py", line 314, in _bootstrap
 (SGLangEngine pid=244598)     self.run()
 (SGLangEngine pid=244598)   File "/workspace/gongrui/.uv_python_cache/cpython-3.12.13-linux-x86_64-gnu/lib/python3.12/multiprocessing/process.py", line 108, in run
 (SGLangEngine pid=244598)     self._target(*self._args, **self._kwargs)
 (SGLangEngine pid=244598)   File "/workspace/gongrui/projects/sglang/python/sglang/srt/entrypoints/http_server.py", line 1932, in launch_server
 (SGLangEngine pid=244598)     _launch_subprocesses(
 (SGLangEngine pid=244598)   File "/workspace/gongrui/projects/sglang/python/sglang/srt/entrypoints/engine.py", line 1053, in _launch_subprocesses
 (SGLangEngine pid=244598)     scheduler_procs, scheduler_pipe_readers = _launch_scheduler_processes(
 (SGLangEngine pid=244598)                                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 (SGLangEngine pid=244598)   File "/workspace/gongrui/projects/sglang/python/sglang/srt/entrypoints/engine.py", line 1006, in _launch_scheduler_processes
 (SGLangEngine pid=244598)     with memory_saver_adapter.configure_subprocess(), numa_utils.configure_subprocess(
 (SGLangEngine pid=244598)          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 (SGLangEngine pid=244598)   File "/workspace/gongrui/.uv_python_cache/cpython-3.12.13-linux-x86_64-gnu/lib/python3.12/contextlib.py", line 137, in __enter__
 (SGLangEngine pid=244598)     return next(self.gen)
 (SGLangEngine pid=244598)            ^^^^^^^^^^^^^^
 (SGLangEngine pid=244598)   File "/workspace/gongrui/projects/slime/amd_env/.venv/lib/python3.12/site-packages/torch_memory_saver/hooks/mode_preload.py", line 27, in
 configure_subprocess
 (SGLangEngine pid=244598)     lib_path = str(get_binary_path_from_package("torch_memory_saver_hook_mode_preload"))
 (SGLangEngine pid=244598)                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 (SGLangEngine pid=244598)   File "/workspace/gongrui/projects/slime/amd_env/.venv/lib/python3.12/site-packages/torch_memory_saver/utils.py", line 65, in get
 _binary_path_from_package
 (SGLangEngine pid=244598)     raise RuntimeError(
 (SGLangEngine pid=244598) RuntimeError: torch_memory_saver: expected exactly one .so matching 'torch_memory_saver_hook_mode_preload_cu12.*.so' (detected CUDA major=12), found
 0: []. This usually means the installed wheel does not match your CUDA runtime.
```

My system installs both cuda and rocm but the GPU is AMD.  I think this may be the root cause because it said cuda is detected. In this case, the linked file should be the amd version "torch_memory_saver_hook_mode_preload.so" and we should not search "torch_memory_saver_hook_mode_preload_cu"

fzyzcjy:
Agent review note: this PR appears to be a follow-up fix for the multi-CUDA wheel change merged in #74 (`cb4fa6b`). That change made CUDA builds emit suffixed extension modules (`_cu12` / `_cu13`) and updated `get_binary_path_from_package()` to always detect a CUDA major and resolve `{stem}_cu{major}.*.so`.

The bug is that ROCm/HIP builds were intentionally left as single-variant, unsuffixed binaries in `setup.py` (`name_suffix = ""` for non-CUDA), but the runtime resolver no longer had a ROCm path after #74. As a result, a ROCm torch install could enter CUDA detection / CUDA-suffixed lookup even though the packaged ROCm binary is named like `{stem}.abi3.so`.

The fix here is logically correct: detect `torch.version.hip` first, and for ROCm/HIP torch resolve the unsuffixed `{stem}.*.so`; otherwise keep the existing CUDA major detection and `_cuXX` lookup path. This preserves the multi-CUDA behavior while restoring the ROCm naming contract.

One small wording caveat: CUDA merged wheels may also contain an unsuffixed cu12 compatibility copy from #74, for hardcoded legacy LD_PRELOAD users. The resolver still deliberately ignores that on CUDA because it searches `_cu{major}`. So the docstring is best read as "ROCm builds use unsuffixed binaries" rather than "only ROCm wheels contain unsuffixed binaries."

PR #78, opened 2026-06-25
Add Intel XPU (Level Zero) backend

siju-samuel:
Adds Intel GPU support alongside the existing CUDA and ROCm backends, providing the same pause/resume (release/re-commit) behavior built natively on Level Zero virtual memory:

```
malloc -> zeVirtualMemReserve + zePhysicalMemCreate + zeVirtualMemMap
pause  -> zeVirtualMemUnmap + zePhysicalMemDestroy   (VA stays reserved)
resume -> zePhysicalMemCreate + zeVirtualMemMap      (same VA)
```

Wired into PyTorch via `torch.xpu.memory.XPUPluggableAllocator` + `torch.xpu.MemPool`, so only `hook_mode="torch"` is supported (LD_PRELOAD preload is CUDA/HIP-specific). Pauseable graph capture, `disable()`, disk backup and `memory_margin_bytes` are not supported on XPU and are rejected explicitly rather than silently ignored.

**C++**

- `macro.h` / `utils.h`: `USE_XPU` platform branch (types, device helpers, host-copy shims so the shared disk-backup path compiles unforked).
- `core.h`: `AllocationMetadata` XPU fields (`XPUAllocExtra`).
- `hardware_xpu_support.{h,cpp}`: Level Zero backend — malloc/free/pause/resume, optional CPU backup, per-device SYCL context prewarm, and byte-accounting (`committed`/`leaked`/`tracked`/`affected_devices`). Uses the SYCL platform default context so VA mappings are visible to torch kernels.
- Transactional error handling: every pause/resume/free step either completes or leaves a recoverable state — a failed `zePhysicalMemDestroy` retains the handle (surfaced as leaked bytes) so resume re-maps that exact handle; `resume` restores the CPU backup before committing; `free` keeps its ownership record until Level Zero confirms release. `tms_pause`/`tms_resume` return non-zero on partial failure and Python raises.
- `core.cpp` / `entrypoint.cpp`: dispatch to `XPUImplementation` under `USE_XPU`. CUDA/ROCm paths unchanged (XPU is an `#elif` before the CUDA `#else`).

**Python**

- `hooks/mode_torch.py`: use `XPUPluggableAllocator` on XPU.
- `entrypoint.py`: `torch.xpu` MemPool/`use_mem_pool`, prewarm, atexit pool cleanup. MemPool key includes the current device so multiple devices in one process are each captured. `pause`/`resume` drain every device the backend will unmap (queried from `tms_xpu_affected_devices`), not just the current one — a missed device risks a `DEVICE_LOST` hang.
- `utils.py` / `binary_wrapper.py`: XPU binary resolution + `tms_xpu_*` signatures.

**Tests**

- `test/test_examples.py` is device-aware: on XPU it runs only `hook_mode='torch'` and skips CUDA/HIP-only examples (`cuda_graph`, `disk_backup`, `multi_device`, `rl_example`, `training_engine`, `nested_region`) with clear reasons.
- `test/examples/multi_device_torch_mode.py`: torch-mode multi-GPU pause/resume (pins each device at alloc time), runs on **both** CUDA and XPU. Preload-mode multi-device coverage stays in the existing `multi_device.py`.
- `test/examples/xpu_scenarios.py`: XPU-only scenarios, each in its own spawned subprocess — fault injection for the resume/cleanup/free transactional paths (asserted via the committed/leaked/tracked counters), multi-device drain, and the `disable()`/`memory_margin_bytes` rejections.
- On Intel XPU: **9 passed, 7 skipped**.

**Build & docs**

- `setup.py`: detect XPU via `icpx`; build with `-fsycl`, link `sycl` + `ze_loader`; build only the torch hook variant. Picks an `icpx` whose `libsycl` major matches the installed `torch+xpu` so the `.so` links a compatible ABI.
- `make build-xpu`: `TMS_PLATFORM=xpu pip install --no-build-isolation .`
- Examples made device-portable via `testing_utils.get_device()`/`empty_cache()`.
- README: Platform Support table + Intel XPU section.

Validated on Intel GPUs: pause/resume release real physical memory and restore usable tensors at the same virtual address. Verified via the driver-independent `tms_xpu_committed_bytes` (sysman free-bytes is frozen on newer Intel drivers), with oneAPI 2025.3 and 2026.0.


fzyzcjy:
Looks great! Feel free to ping me when it is ready; if I do not see the comment, also feel free to ping me in SGLang slack

siju-samuel:
@fzyzcjy please review and merge. Thanks in advance
Verified with Intel OneAPI version 2025.3 and 2026.0

siju-samuel:
@fzyzcjy Somehow i was not able to find you in SGLang slack.
 Could you please have a look at this PR when you get a chance?

fzyzcjy:
Let me review now. SGLang slack is here:

image

csrc/entrypoint.cpp
```diff
@@ -63,6 +63,9 @@ void *tms_torch_malloc(ssize_t size, int device, cudaStream_t stream) {
               << " size=" << size << " device=" << device << " stream=" << stream
               << std::endl;
 #endif
+    // The torch MemPool allocator is pool-scoped: this hook only fires for
```
fzyzcjy: nit: would be great to make comments as dense as possible

csrc/entrypoint.cpp
```diff
@@ -76,7 +79,15 @@ void tms_torch_free(void *ptr, ssize_t ssize, int device, cudaStream_t stream) {
               << " ptr=" << ptr << " ssize=" << ssize << " device=" << device << " stream=" << stream
               << std::endl;
 #endif
+#if !defined(USE_XPU)
```
fzyzcjy, line 93: quick question: why XPU skips this check

test/examples/xpu_multi_device.py
fzyzcjy, line 1: quick question: is it possible to make it not a xpu-only test, but also run on other platforms (e.g. cuda)

torch_memory_saver/hooks/mode_torch.py
```diff
@@ -1,15 +1,30 @@
...
+import torch
+
 from torch_memory_saver.hooks.base import HookUtilBase
 from torch_memory_saver.utils import get_binary_path_from_package
-from torch.cuda.memory import CUDAPluggableAllocator
 
 logger = logging.getLogger(__name__)
 
 
+def _is_xpu() -> bool:
```
fzyzcjy: nit: this func appears multi times, maybe put in utils.py

torch_memory_saver/entrypoint.py
```diff
@@ -96,27 +100,64 @@ def _ensure_initialized(self):
...
+                )
+            self._mem_pools = defaultdict(
+                lambda: torch.xpu.MemPool(allocator=self._hook_util.get_allocator())
+            )
+            self._use_mem_pool = torch.xpu.use_mem_pool
+        else:
+            self._mem_pools = defaultdict(
+                lambda: torch.cuda.MemPool(allocator=self._hook_util.get_allocator())
+            )
+            self._use_mem_pool = torch.cuda.use_mem_pool
```
fzyzcjy: would be great to reduce code dup, e.g. 

mod = torch.get_device_module()
mem_pools = mod.MemPool(...)

torch_memory_saver/entrypoint.py
```diff
@@ -96,27 +100,64 @@ def _ensure_initialized(self):
...
         # (For hook_mode=torch we may not need it, but currently our primary usage is hook_mode=preload, thus we do this for simplicity)
-        mem_pool = self._mem_pools[(tag, enable_cpu_backup)]
-        with torch.cuda.use_mem_pool(mem_pool):
+        #
+        # A torch MemPool is bound to the device that is current when it is
+        # created; reusing it for an allocation on another device silently
+        # bypasses the custom allocator. So the pool key must include the device
+        # to support multiple devices in one process (e.g. several TP ranks, or
+        # tests that touch multiple GPUs).
+        key = (tag, enable_cpu_backup, self._current_device())
```
fzyzcjy: is this XPU only, or does CUDA also have this issue? if latter, this looks like a good fix, could you please create another PR (w/ a test to protect it), and I will merge that one first, to ensure PRs are more atomic

siju-samuel:
Thanks @fzyzcjy 
image
Somehow SGL wont show up your name.
Hope you got this message in slack
image

torch_memory_saver/entrypoint.py
```diff
@@ -96,27 +100,64 @@ def _ensure_initialized(self):
...
+        mem_pool = self._mem_pools[key]
+        with self._use_mem_pool(mem_pool):
             with self._with_region_config(tag=tag, enable_cpu_backup=enable_cpu_backup):
                 yield
 
+    def _current_device(self) -> int:
+        if self._is_xpu:
+            return torch.xpu.current_device()
+        return torch.cuda.current_device()
```
fzyzcjy: what about torch.get_device_module().current_device()

(same for many other places)

README.md
```diff
@@ -85,6 +85,86 @@ torch_memory_saver.hook_mode = "torch"
 
 Please refer to `rl_example.py` for details.
 
+## Platform Support
```
fzyzcjy, line 88: nit: do not use bold text (`**sth**`)

README.md
```diff
@@ -85,6 +85,86 @@ torch_memory_saver.hook_mode = "torch"
...
+from torch_memory_saver import torch_memory_saver
+
+torch_memory_saver.hook_mode = "torch"  # required on XPU
+
+with torch_memory_saver.region(tag="weights"):
+    x = torch.full((1_000_000_000,), 100, dtype=torch.uint8, device="xpu")
+
+torch_memory_saver.pause("weights")    # physical memory returned to the device
+torch_memory_saver.resume("weights")   # re-committed at the same virtual address
+```
```
fzyzcjy: nit: be brief, e.g. this part may be removed

setup.py
```diff
@@ -33,13 +46,127 @@ def _find_platform_home(platform):
     return home
 
 
+def _find_icpx():
```
fzyzcjy: nit: maybe write sth

```
# ====================== For XPU ======================
<your code here>
# ===================================================
```

for the new functionsto make it slightly more organized

fzyzcjy:
@siju-samuel just `@Tom` in sgl slack; I do not recv @ for that message indeed

csrc/entrypoint.cpp
siju-samuel: `tms_torch_free `is the MemPool free callback, so it only fires for tensors we allocated in a region(). But those tensors are freed by Python GC, which often runs after the region() block exits. so `is_interesting_region() `is already `False `and the assert would abort on a valid free. The pointer is always one we manage, so `free(ptr) ` handles it directly. CUDA/ROCm kept unchanged (`#if !defined(USE_XPU)`).

torch_memory_saver/hooks/mode_torch.py
siju-samuel: Done

test/examples/xpu_multi_device.py
siju-samuel: test/examples/multi_device_torch_mode.py changed to device-agnostic

torch_memory_saver/entrypoint.py
siju-samuel: https://github.com/fzyzcjy/torch_memory_saver/pull/81

csrc/entrypoint.cpp
fzyzcjy: hmm a bit confused, do you mean XPU is doing sth different, or is it a pre-existing bug on CUDA

fzyzcjy (commented):
generally LGTM, will have a deeper look (and w/ agents) after prereq pr is merged

csrc/entrypoint.cpp
siju-samuel: Right. I guess CUDA's default and tested path is preload, whose cudaFree has no such assert, so CUDA in practice is unaffected. 
But XPU is torch-mode-only, where the free is the MemPool callback and Python GC routinely frees the tensor after the region() block exits and hits this assert. That's why I guard the XPU.

fzyzcjy:
hi, could you plz resolve merge conflicts (and I will have a final review ASAP having time)

siju-samuel:
> hi, could you plz resolve merge conflicts (and I will have a final review ASAP having time)

Thanks. I have rebased and resolved the merge conflicts.

fzyzcjy:
looks great, busy recently and will review ASAP when having time

fzyzcjy (changes requested):

torch_memory_saver/entrypoint.py
```diff
@@ -156,25 +180,44 @@ def disable(self, dispose_mem_pool_after_use: bool = True):
 
         self._binary_wrapper.cdll.tms_set_interesting_region(False)
         try:
-            # We can either reuse the pool or delete it immediately, and we implement the latter currently since Slime uses it.
-            # About why we need a pool: https://github.com/fzyzcjy/torch_memory_saver/pull/20#issuecomment-3047099047
-            pool = torch.cuda.MemPool()
-            with torch.cuda.use_mem_pool(pool):
+            if self._is_xpu:
```
fzyzcjy: **[P1] `disable()` still uses this custom XPU pool and aborts on the next allocation.** This branch sets `interesting_region=False` and then yields without switching away from the outer custom `MemPool`. The next allocation still reaches `tms_torch_malloc`, whose unconditional `SIMPLE_CHECK(thread_local_config.is_interesting_region())` calls `exit(1)`. The comment refers to a passthrough path, but that path was removed, and the only `disable()` coverage is skipped on XPU. Please implement a real allocator bypass/default-pool switch, or explicitly reject and document this API instead of leaving a process-killing path.

torch_memory_saver/entrypoint.py
```diff
@@ -156,25 +180,44 @@ def disable(self, dispose_mem_pool_after_use: bool = True):
...
 
     def pause(self, tag: Optional[str]):
+        if self._is_xpu:
+            # Must not unmap physical pages while kernels are still touching
+            # them: zeVirtualMemUnmap on in-flight pages hangs the device, so
+            # drain the device first. We sync the CURRENT device only; like the
+            # CUDA backend, cross-device synchronization is the caller's job.
+            # Call pause()/resume() with the region's device current -- the
+            # natural one-process-per-rank (one device) deployment already does.
+            self._device_module.synchronize()
```
fzyzcjy: **[P1] This synchronizes only the current device, but native `xpu_pause()` unmaps every matching allocation on every device.** The multi-device test has to manually synchronize both devices before calling `pause()`, while normal callers are not protected. If a non-current device still has an in-flight kernel, its VA can be unmapped and trigger the `DEVICE_LOST`/hang described above. Please synchronize every affected device inside the backend, or make pause/resume explicitly device-scoped and only touch that device's metadata.

csrc/hardware_xpu_support.cpp
```diff
@@ -0,0 +1,460 @@
...
+
+      ze_context_handle_t ze_ctx = metadata.xpu.ze_ctx;
+      ze_device_handle_t ze_dev = metadata.xpu.ze_dev;
+      size_t aligned = metadata.aligned_size;
+
+      ze_physical_mem_desc_t pdesc = {};
+      pdesc.stype = ZE_STRUCTURE_TYPE_PHYSICAL_MEM_DESC;
+      pdesc.size = aligned;
+      ze_physical_mem_handle_t phys = {};
+      ze_result_t rc = zePhysicalMemCreate(ze_ctx, ze_dev, &pdesc, &phys);
```
fzyzcjy: **[P1] A create/map failure is silently treated as a successful `resume()`.** Both failure branches only log and `continue`, while the C entrypoint and Python API return no status. A tag can therefore be half-resumed, with some tensors still unmapped; the next access can device-fault even though `resume()` returned normally. Please make the transition transactional, roll back partial work, propagate an error to Python, and add failure-injection coverage for create/map failures.

csrc/hardware_xpu_support.cpp
```diff
@@ -0,0 +1,460 @@
...
+                           ZE_MEMORY_ACCESS_ATTRIBUTE_READWRITE);
+      if (rc != ZE_RESULT_SUCCESS) {
+        zePhysicalMemDestroy(ze_ctx, phys);
+        XPU_ERR("resume zeVirtualMemMap failed: 0x" << std::hex << rc);
+        continue;
+      }
+      metadata.xpu.ze_phys = phys;
+      metadata.state = AllocationState::ACTIVE;
+
+      if (metadata.enable_cpu_backup && metadata.cpu_backup) {
```
fzyzcjy, line 388: **[P1] A restore-copy failure permanently discards the only CPU backup.** The allocation is marked `ACTIVE` before this copy; an exception is only logged, and the code then unconditionally frees `cpu_backup`. The caller observes a successful resume with undefined tensor contents and no retry copy. Keep the allocation/backup in a recoverable state until the copy succeeds, roll back the new mapping on failure, and surface the error.

csrc/hardware_xpu_support.cpp
```diff
@@ -0,0 +1,460 @@
...
+    std::unordered_map<void *, AllocationMetadata> &allocation_metadata,
+    std::mutex &allocator_metadata_mutex) {
+  AllocationMetadata metadata;
+  {
+    const std::lock_guard<std::mutex> lock(allocator_metadata_mutex);
+    auto it = allocation_metadata.find(ptr);
+    if (it == allocation_metadata.end())
+      return cudaErrorInvalidDevicePointer;
+    metadata = it->second;
+    allocation_metadata.erase(it);
```
fzyzcjy: **[P1] `xpu_free()` drops the only ownership record before any Level Zero cleanup succeeds.** The subsequent unmap/destroy/free return values are all ignored, so any failure permanently leaks physical memory or VA and cannot be retried; an unmap failure is followed by destroying a still-mapped physical handle, despite `xpu_pause()` explicitly avoiding that unsafe sequence. Please retain explicit ownership/state until every cleanup step succeeds and propagate cleanup failures to the allocator.

csrc/hardware_xpu_support.cpp
```diff
@@ -0,0 +1,460 @@
...
+        XPU_ERR("pause zeVirtualMemUnmap failed: 0x" << std::hex << rc_unmap
+                                                     << "; keeping ACTIVE");
+        continue;
+      }
+      ze_result_t rc_destroy =
+          zePhysicalMemDestroy(ze_ctx, metadata.xpu.ze_phys);
+      if (rc_destroy != ZE_RESULT_SUCCESS)
+        XPU_ERR("pause zePhysicalMemDestroy failed: 0x" << std::hex
+                                                        << rc_destroy);
+      metadata.xpu.ze_phys = {};
```
fzyzcjy, line 319: **[P1] Do not clear this handle or mark the allocation paused when `zePhysicalMemDestroy` failed.** On failure the old physical allocation remains owned, but these state updates make it unreachable; the next resume creates another handle. `tms_xpu_committed_bytes()` then reports zero because it counts only `ACTIVE` state, so the new tests mask the real leak. Preserve a recoverable intermediate/error state, propagate the failure, and validate cleanup with failure injection rather than only the self-reported state counter.

csrc/core.cpp
```diff
@@ -14,6 +14,9 @@ cudaError_t TorchMemorySaver::malloc(void **ptr, CUdevice device, size_t size, c
 #if TMS_ROCM_LEGACY_CHUNKED
     return ROCmHIPImplementation::rocm_malloc(ptr, device, size, tag, enable_cpu_backup, allocation_metadata_, allocator_metadata_mutex_);
 
+#elif defined(USE_XPU)
```
fzyzcjy, line 36: **[P2] This early XPU return silently bypasses the public `memory_margin_bytes` contract.** The margin check is below this branch, while the Python setter remains available on XPU without warning. A caller relying on the configured safety margin can still exhaust the device. Please implement the check in the XPU backend or reject this setting explicitly.

setup.py
```diff
@@ -65,15 +84,37 @@ def build_extensions(self):
...
+        if self.platform == "xpu":
+            # Set icpx (Intel oneAPI SYCL compiler) for XPU, ABI-matched to torch.
+            icpx = _resolve_xpu_icpx()
+            self.compiler.set_executable("compiler_so", icpx)
+            self.compiler.set_executable("compiler_cxx", icpx)
+            self.compiler.set_executable("linker_so", f"{icpx} -shared")
+            for ext in self.extensions:
+                # -std=c++17 explicitly: the backend uses structured bindings
+                # (auto &[ptr, metadata]); don't rely on icpx's default standard.
+                ext.extra_compile_args = ['-std=c++17', '-fPIC', '-fsycl', '-fsycl-targets=spir64']
```
fzyzcjy: **[P2] This assignment overwrites the common `['-std=c++17', '-O3']` arguments.** The entire XPU allocator backend is therefore built at icpx's default optimization level. Please append the SYCL-specific flags to the existing `extra_compile_args` instead of replacing them, and cover the final compile command in a build test.

fzyzcjy (changes requested):
> 🤖 **Agent-authored inline review.** This review was generated and submitted autonomously by a coding agent on behalf of the repository owner. The findings below were independently checked across semantic, lifecycle, architecture, performance, and adversarial passes. Please point out anything inaccurate.

Requesting changes for the blocking findings attached to the exact diff lines. This was a source-level review; no Intel XPU hardware was available locally.

csrc/entrypoint.cpp
```diff
@@ -76,7 +79,15 @@ void tms_torch_free(void *ptr, ssize_t ssize, int device, cudaStream_t stream) {
...
+    // Upstream CUDA/ROCm behavior: assert we are in an interesting region.
     SIMPLE_CHECK(thread_local_config.is_interesting_region(), "only support interesting region");
+#endif
+    // XPU is torch-mode only with a pool-scoped allocator, so this hook fires
+    // exactly for the region() tensors we allocated. Those are freed when they
+    // go out of scope, which can happen AFTER the region context has exited
+    // (is_interesting_region() back to false) -- hence no assert on XPU. The
+    // pointer is always VMM-managed here, so free() handles it directly.
```
fzyzcjy: nit: would be great to make this and other comments dense

test/examples/multi_device_torch_mode.py
```diff
@@ -1,69 +1,74 @@
...
-_ALLOC = 100_000_000  # ~100 MB uint8
-_FREED = 80_000_000   # expected drop/restore, with slack
+_ALLOC_BYTES = 256 * 1024 * 1024 * 4  # 256M float32 = 1 GiB
+_FREED_GIB = 0.8  # expected drop/restore, with slack for alignment/noise
 
 
 def run(hook_mode: str):
-    torch_memory_saver.hook_mode = hook_mode
-    logging.basicConfig(level=logging.DEBUG, stream=sys.stdout)
+    assert hook_mode == "torch", "multi-device torch-mode test requires hook_mode='torch'"
```
fzyzcjy, line 29: quick question: seems like we are weakening the test, is it possible to also test non-torch mode

torch_memory_saver/entrypoint.py
```diff
@@ -96,32 +97,55 @@ def _ensure_initialized(self):
...
+                "allocator); preload/LD_PRELOAD is CUDA-only."
+            )
         self._hook_mode = hook_mode
         self._hook_util = HookUtilBase.create(hook_mode=hook_mode)
         self._binary_wrapper = BinaryWrapper(path_binary=self._hook_util.get_path_binary())
-        self._mem_pools = defaultdict(lambda: torch.cuda.MemPool(allocator=self._hook_util.get_allocator()))
+
+        # torch.cuda / torch.xpu share the MemPool / use_mem_pool / current_device
+        # / synchronize API; resolve the right module once instead of branching
+        # per platform.
```
fzyzcjy: nit: rm this comment

torch_memory_saver/entrypoint.py
```diff
@@ -96,32 +97,55 @@ def _ensure_initialized(self):
...
+            # Prewarm per-device SYCL contexts before registering the allocator.
+            if hasattr(self._binary_wrapper.cdll, "tms_xpu_prewarm_devices"):
+                self._binary_wrapper.cdll.tms_xpu_prewarm_devices(
+                    self._device_module.device_count()
+                )
+
+        self._mem_pools = defaultdict(
+            lambda: self._device_module.MemPool(allocator=self._hook_util.get_allocator())
+        )
+        self._use_mem_pool = self._device_module.use_mem_pool
```
fzyzcjy: nit: rm this line, i.e. inline it

fzyzcjy (approved):
LGTM w/ nits (below are human reviews; above are agent reviews as serve as references not mandatory)

fzyzcjy:
(feel free to ping me when ready)

siju-samuel:
> (feel free to ping me when ready)

There were merge conflicts due to https://github.com/fzyzcjy/torch_memory_saver/pull/80. I'm validating the changes to make sure everything still works correctly after resolving them.

siju-samuel:
image

@fzyzcjy I've completed the changes from my side and re-verified everything after resolving the merge conflicts. The PR is ready for review again. When you have a chance, could you please take another look and help merge it? Thanks!

test/examples/multi_device_torch_mode.py
siju-samuel: Non-torch (preload) multi-device is still covered by the existing multi_device.py (preload, CUDA-only); this new file is the torch-mode counterpart, so preload coverage isn't lost. On XPU preload can't run.. the backend asserts hook_mode="torch" since preload hooks cudaMalloc.

csrc/entrypoint.cpp
```diff
@@ -141,17 +147,70 @@ void set_memory_margin_bytes(uint64_t value) {
...
-    TorchMemorySaver::instance().resume(tag_str);
+    return static_cast<int>(TorchMemorySaver::instance().resume(tag_str));
 }
 
 uint8_t* tms_get_cpu_backup_pointer(const uint8_t* gpu_ptr, uint64_t size) {
     return TorchMemorySaver::instance().get_cpu_backup_pointer(gpu_ptr, size);
 }
+
+#if defined(USE_XPU)
+// Pre-warm per-device SYCL contexts from Python init: creating a sycl::context
```
fzyzcjy: nit: only have absolutely necessary comments, i.e. remove most of them in the pr

torch_memory_saver/entrypoint.py
```diff
@@ -185,15 +241,67 @@ def disable(self, dispose_mem_pool_after_use: bool = True):
...
+                raise RuntimeError(
+                    f"torch_memory_saver.resume(tag={tag!r}) partially failed "
+                    f"(code {ret}); some allocations remain paused. Retry after "
+                    "resolving the failure (e.g. free device memory)."
+                )
+            # Settle each re-mapped device's stream so the driver finishes paging/TLB
+            # work before user kernels touch the remapped addresses.
+            self._sync_affected_devices(tag)
+
+    def _sync_affected_devices(self, tag: Optional[str]):
```
fzyzcjy: hmm why add some functions that are not guarded under xpu

torch_memory_saver/entrypoint.py
```diff
@@ -185,15 +233,66 @@ def disable(self, dispose_mem_pool_after_use: bool = True):
...
+
+    def _xpu_sync_affected_devices(self, tag: Optional[str]):
+        # XPU only (callers guard on _is_xpu). Drain every device the backend will
+        # unmap/remap for `tag` (a missed one risks a DEVICE_LOST hang);
+        # synchronize(device) preserves the current device.
+        for device in self._xpu_affected_devices(tag):
+            self._device_module.synchronize(device)
+
+    def _xpu_affected_devices(self, tag: Optional[str]) -> list[int]:
+        """Device ids the XPU backend will unmap/remap for `tag` (authoritative).
```
fzyzcjy, line 254: nit: would be great to manually cleanup the code a bit when resolving previous comments, e.g. as mentioned earlier, avoid lengthy comments

siju-samuel: Thanks. 
Sure. im removing all the comments. Will ping you for review once im done.

siju-samuel:
Thanks for the review and suggestions, 
@fzyzcjy  Could you please take another look when you have a chance?
I've removed all the unnecessary comments.

fzyzcjy:
sure, will check asap

fzyzcjy (approved):

PR #79, opened 2026-07-01
feat(hooks): add pluggable hook modes

### LINKED ISSUE #77: [RFC] Pluggable memory-saver implementation in torch_memory_saver
## Summary

This RFC proposes two small extension points in `torch_memory_saver` so that an external allocator can provide its own memory-saver implementation, selected through the existing `hook_mode` mechanism. The motivating consumer is an out-of-process GPU memory service that owns GPU memory behind stable virtual addresses, but the asks are generic. [This](https://github.com/sgl-project/sglang/issues/27310) is a RFC in SGLang that relies on the changes proposed on this RFC.

## Motivation

`torch_memory_saver` pauses and resumes tensors by releasing physical memory while preserving the virtual address, then re-allocating physical memory on resume. That same VA-stable unmap/remap is valuable to allocators other than the bundled in-process one — for example an out-of-process GPU memory service that owns the physical handles in a separate process so they can be shared across processes on one GPU. This lets a process restart and reattach to resident weights instead of reloading them, and (with a standby) supports shadow-engine failover.

Such an allocator wants to reuse the existing `region` / `pause` / `resume` API but supply its own implementation: its `pause` releases its physical backing and drops its external session, its `resume` reconnects and remaps, and its allocations route to its own allocator rather than the bundled one. The `region` / `pause` / `resume` abstraction fits exactly; the obstacle is that the implementation behind it is private and the set of hook modes that selects it is closed, so there is no supported way to provide an alternative. This RFC proposes the minimal surface to close that gap.

### 1\. Make the hook-mode set extensible

**Why.** The implementation that provides `region(tag, enable_cpu_backup)`, `pause(tag)`, `resume(tag)`, and `cuda_graph(...)` is selected by `hook_mode` and built inside `_ensure_initialized()`, but `HookMode` is a closed literal (`preload`, `torch`) resolved by a hardcoded factory. There is no supported way to select an alternative implementation.

**Change.** Allow a custom `hook_mode` to be registered together with the implementation it builds, selected through the existing `hook_mode` property:

```py
import torch_memory_saver
torch_memory_saver.register_hook_mode("gms", my_impl_factory, uses_preload=False)
torch_memory_saver.torch_memory_saver.hook_mode = "gms"   # existing setter, before init
```

The factory is invoked lazily where `_ensure_initialized` builds the implementation, so it can capture the device and runtime configuration the implementation needs, and the implementation must satisfy the `region` / `pause` / `resume` / `cuda_graph` / `enabled` contract `_TorchMemorySaverImpl` already defines.

Note that a registered mode swaps the whole implementation, which is a layer above where the built-in `hook_mode` selects a `HookUtil` for the single bundled implementation (see open questions).

### 2\. Make `configure_subprocess` hook-mode-aware

**Why.** `configure_subprocess()` performs the `LD_PRELOAD` setup that the `preload` hook mode requires for spawned child processes. An implementation that does not use the preload path must be able to skip it.

**Change.** Have `configure_subprocess` no-op for modes that do not use the preload setup. The signal must come from the mode itself (e.g. the `uses_preload=False` given at registration, defaulting to `True` so the built-in modes are unchanged), because `configure_subprocess` runs in the parent before the implementation is constructed.

## Out of scope

The external implementation itself (its `region` / `pause` / `resume` logic, allocator, sessions, and locks) lives in the consumer, not here.

## Open questions

- Registration shape: a `register_hook_mode(name, factory, uses_preload=...)` call, or letting `HookMode` accept a class/factory directly?  
- Layering: a registered mode swaps the whole implementation, above where `hook_mode` currently selects a `HookUtil` for the one bundled implementation. Is overloading `hook_mode` for this acceptable, or should implementation selection be a distinct concept?  
- Selection timing: the implementation is built lazily on first use, so a mode must be selected before then. Should selecting after initialization raise, given import order is not always controlled (e.g. under multiprocessing spawn)?

CC: @fzyzcjy

galletas1712:
Closes #77 (pluggable memory-saver implementations) — the `torch_memory_saver` piece of the GPU Memory Service integration proposed in sgl-project/sglang#27310.

- `register_hook_mode(name, factory, uses_preload=...)` registers a lazily constructed custom implementation, selected through the existing `hook_mode` setter. Factories are process-global and resolved on first use; the builtin `preload`/`torch` modes are registered the same way and behave exactly as before.
- The package-level `configure_subprocess()` now consults the selected mode's `uses_preload` flag and skips the `LD_PRELOAD` setup for modes that do not use the preload hook (both builtins keep it, preserving current behavior).
- Selecting an unknown mode, re-registering a name, or using a mode never registered in the current process (e.g. a subprocess that skipped registration) fails with a descriptive error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,20 @@ The mode can be chosen by:
 torch_memory_saver.hook_mode = "torch"
 ```
 
+Custom implementations can register a lazy factory:
+
+```python
+from torch_memory_saver import torch_memory_saver, register_hook_mode
+
+register_hook_mode("custom", lambda: CustomMemorySaver(), uses_preload=False)
+torch_memory_saver.hook_mode = "custom"
+```
+
+The implementation must provide `region`, `cuda_graph`, `pause`, and `resume`
+(plus any other facade method you call). Select the mode before first use;
+`configure_subprocess()` skips the `LD_PRELOAD` setup for modes registered
+with `uses_preload=False`.
+
 ### Example of RL with CUDA Graph
 
 Please refer to `rl_example.py` for details.
--- a/torch_memory_saver/__init__.py
+++ b/torch_memory_saver/__init__.py
@@ -1,5 +1,17 @@
-from .entrypoint import TorchMemorySaver
-from .hooks.mode_preload import configure_subprocess
+from contextlib import contextmanager
+
+from .entrypoint import TorchMemorySaver, register_hook_mode
+from .hooks import mode_preload as _mode_preload
 
 # Global singleton
 torch_memory_saver = TorchMemorySaver()
+
+
+@contextmanager
+def configure_subprocess():
+    if not torch_memory_saver._uses_preload():
+        yield
+        return
+
+    with _mode_preload.configure_subprocess():
+        yield
--- a/torch_memory_saver/entrypoint.py
+++ b/torch_memory_saver/entrypoint.py
@@ -6,7 +6,7 @@ import logging
 import os
 from collections import defaultdict
 from contextlib import contextmanager
-from typing import Optional
+from typing import Any, Callable, Dict, NamedTuple, Optional
 import torch
 
 from .binary_wrapper import BinaryWrapper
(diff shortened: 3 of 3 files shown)


fzyzcjy:
(replied in https://github.com/fzyzcjy/torch_memory_saver/issues/77)

PR #80, opened 2026-07-04
feat: disk backup backend, spill paused allocations to node-local disk

Zhichenzzz:
An `enable_disk_backup` mode alongside `enable_cpu_backup`: paused allocations are spilled to node-local files instead of a pinned CPU buffer.

## Why
`enable_cpu_backup` holds a full pinned host copy of every paused allocation. For large training actors (e.g. RL colocate) that can exceed host RAM. Disk backup streams **GPU → fixed pinned staging → file (`pwrite`)** and back on resume, so the pinned host buffer stays bounded to one staging chunk regardless of the total offloaded size — for the case where even CPU offload does not fit.

## Design
All the disk machinery lives in a standalone `DiskBackend` class (`csrc/disk_backend.{h,cpp}`), not mixed into `TorchMemorySaver`. Each allocation carries a small `DiskBackupSlot` (just the file path); `pause()`/`resume()` delegate to `DiskBackend::offload`/`onload`.

- **Atomic, unique files.** Files are created with `mkostemp` (`O_EXCL` + `O_CLOEXEC`), not a predictable `getpid()+counter` path — two containers sharing a PID namespace and a node-local mount could otherwise open the same inode and corrupt each other.
- **Bounded fds.** The fd is opened per transfer and closed after. PyTorch's MemPool caches allocator blocks, so a persistent per-block fd would grow with fragmentation and eventually hit `RLIMIT_NOFILE`; a transient fd keeps it flat.
- **Bounded disk.** The file is `posix_fallocate`d once (with an `ftruncate` fallback for mounts that don't support it) and overwritten in place every pause, so disk usage never grows across pause/resume cycles. Files are unlinked on free.
- **Cache control.** After each transfer, `fsync` + `posix_fadvise(POSIX_FADV_DONTNEED)` drops the file's page cache so cached bytes don't accumulate. A tmpfs/ramfs target is warned about — there the file itself lives in RAM.
- **Native mutual exclusion.** cpu/disk exclusion is enforced at the native `malloc` boundary, not only via a Python `assert` (which `-O` strips and direct C-API / env-var use bypasses).

## API
- `region(enable_disk_backup=True)`; the target directory is **process-global** (`TMS_DISK_BACKUP_DIR` or `set_disk_backup_dir()`), not region-scoped.
- C: `tms_{set,get}_enable_disk_backup`, `tms_set_disk_backup_dir`
- env: `TMS_INIT_ENABLE_DISK_BACKUP`, `TMS_DISK_BACKUP_DIR`, `TMS_DISK_BACKUP_CHUNK_MB` (default 256)

## Scope / notes
- Host-RAM bound is over the **pinned staging buffer + dropped page cache**; it is not a bound to zero on a RAM-backed mount. Point the dir at real disk.
- `get_cpu_backup_pointer` returns null for disk-backed allocations (no CPU-resident copy).
- ROCm 6.x legacy chunked path does not support disk backup and now hard-fails there rather than silently taking the wrong branch.
- GDS/cuFile is intentionally omitted in v1 (nvidia-fs absent in our target envs); the pinned-staging path gives the same bounded-buffer guarantee with no extra deps and can gain a cuFile fast-path later behind an import guard.

## Testing
**Unit / examples** — full suite green on 8×H200 (CUDA 13):
```
$ python -m pytest test/test_examples.py -q
...............
15 passed in 104.75s
```
`test/examples/disk_backup.py` forces a 1 MiB chunk so the tensors span ~13 chunks with a non-even tail, and compares the restored tensors byte-for-byte against an independent CPU oracle across 3 pause/resume cycles while asserting the on-disk footprint stays constant.

**End-to-end** — real miles RL run, Qwen3.5-35B-A3B, 8×H200, colocate, EP=8, `--offload-train-target=disk`:
- Ray job **SUCCEEDED**; all 3 rollouts × 2 train steps `outcome=NORMAL`.
- offload↔resume cycled cleanly every step: per-GPU used memory dropped to ~8–17 GB on offload and returned to ~64–108 GB on resume, 8 ranks consistent, tensors restored correctly (per-step logprob abs-diff stable).
- Backup files created as `tms_XXXXXX` (mkostemp); no `Too many open files`, no OOM, no mutual-exclusion trips across the whole run.

### DIFF
--- a/csrc/core.cpp
+++ b/csrc/core.cpp
@@ -3,73 +3,8 @@
 #include "macro.h"
 #include "api_forwarder.h"
 
-#include <algorithm>
-#include <cstdlib>
-#include <fcntl.h>
-#include <unistd.h>
-#include <sys/types.h>
-
-TorchMemorySaver::TorchMemorySaver() {
-    const char* dir = std::getenv("TMS_DISK_BACKUP_DIR");
-    disk_backup_dir_ = (dir != nullptr) ? std::string(dir) : std::string("/tmp");
-
-    const char* chunk_mb = std::getenv("TMS_DISK_BACKUP_CHUNK_MB");
-    size_t mb = (chunk_mb != nullptr) ? std::strtoull(chunk_mb, nullptr, 10) : 256;
-    if (mb == 0) mb = 256;
-    disk_chunk_bytes_ = mb * 1024ull * 1024ull;
-}
-
-void TorchMemorySaver::ensure_disk_staging_() {
-    if (disk_staging_ == nullptr) {
-        CUDA_ERROR_CHECK(cudaMallocHost(&disk_staging_, disk_chunk_bytes_));
-        SIMPLE_CHECK(disk_staging_ != nullptr, "failed to allocate pinned disk staging buffer");
-    }
-}
-
-void TorchMemorySaver::disk_offload_(void* ptr, AllocationMetadata& metadata) {
-    ensure_disk_staging_();
-
-    if (metadata.disk_fd < 0) {
-        metadata.disk_path = disk_backup_dir_ + "/tms_" + std::to_string((long)getpid())
-            + "_" + std::to_string(disk_backup_counter_++) + ".bin";
-        metadata.disk_fd = open(metadata.disk_path.c_str(), O_RDWR | O_CREAT, 0600);
-        SIMPLE_CHECK(metadata.disk_fd >= 0, "failed to open disk backup file");
-        int fret = posix_fallocate(metadata.disk_fd, 0, (off_t) metadata.size);
-        SIMPLE_CHECK(fret == 0, "posix_fallocate on disk backup file failed");
-    }
-
-    size_t offset = 0;
-    while (offset < metadata.size) {
-        size_t n = std::min(disk_chunk_bytes_, metadata.size - offset);
-        CUDA_ERROR_CHECK(cudaMemcpy(disk_staging_, (uint8_t*) ptr + offset, n, cudaMemcpyDeviceToHost));
-        size_t written = 0;
-        while (written < n) {
-            ssize_t w = pwrite(metadata.disk_fd, (uint8_t*) disk_staging_ + written, n - written, (off_t)(offset + written));
-            SIMPLE_CHECK(w > 0, "pwrite to disk backup file failed");
-            written += (size_t) w;
-        }
-        offset += n;
-    }
-}
-
-void TorchMemorySaver::disk_reload_(void* ptr, AllocationMetadata& metadata) {
-    ensure_disk_staging_();
-    SIMPLE_CHECK(metadata.disk_fd >= 0, "disk backup fd invalid on resume (was it ever paused?)");
-
-    size_t offset = 0;
-    while (offset < metadata.size) {
-        size_t n = std::min(disk_chunk_bytes_, metadata.size - offset);
-        size_t got = 0;
-        while (got < n) {
-            ssize_t r = pread(metadata.disk_fd, (uint8_t*) disk_staging_ + got, n - got, (off_t)(offset + got));
-            SIMPLE_CHECK(r > 0, "pread from disk backup file failed");
-            got += (size_t) r;
-        }
-        CUDA_ERROR_CHECK(cudaMemcpy((uint8_t*) ptr + offset, disk_staging_, n, cudaMemcpyHostToDevice));
-        offset += n;
-    }
-    // fd stays open for the next pause.
-}
+TorchMemorySaver::TorchMemorySaver()
+    : disk_backend_(compute_disk_backup_dir_from_env(), compute_disk_chunk_bytes_from_env()) {}
 
 TorchMemorySaver &TorchMemorySaver::instance() {
     static TorchMemorySaver instance;
@@ -77,7 +12,12 @@ TorchMemorySaver &TorchMemorySaver::instance() {
 }
 
 cudaError_t TorchMemorySaver::malloc(void **ptr, CUdevice device, size_t size, const std::string& tag, const bool enable_cpu_backup, const bool enable_disk_backup) {
+    // Enforce here, not only in the Python layer: an assert is stripped under
+    // python -O and bypassed by direct C-API / env-var use.
+    SIMPLE_CHECK(!(enable_cpu_backup && enable_disk_backup),
+                 "cpu_backup and disk_backup are mutually exclusive");
 #if TMS_ROCM_LEGACY_CHUNKED
+    SIMPLE_CHECK(!enable_disk_backup, "disk backup is not supported on the ROCm 6.x legacy chunked path");
     return ROCmHIPImplementation::rocm_malloc(ptr, device, size, tag, enable_cpu_backup, allocation_metadata_, allocator_metadata_mutex_);
 
 #else
@@ -111,7 +51,7 @@ cudaError_t TorchMemorySaver::malloc(void **ptr, CUdevice device, size_t size, c
         allocation_metadata_.emplace(
             *ptr,
             AllocationMetadata{size, device, tag, AllocationState::ACTIVE, enable_cpu_backup, nullptr,
-                               enable_disk_backup, -1, std::string(), allocHandle}
+                               enable_disk_backup, DiskBackupSlot{}, allocHandle}
         );
     }
 
@@ -153,10 +93,8 @@ cudaError_t TorchMemorySaver::free(void *ptr) {
         metadata.cpu_backup = nullptr;
     }
 
-    if (metadata.enable_disk_backup && metadata.disk_fd >= 0) {
-        close(metadata.disk_fd);
-        unlink(metadata.disk_path.c_str());
-        metadata.disk_fd = -1;
+    if (metadata.enable_disk_backup) {
+        disk_backend_.release(metadata.disk);
     }
 
 #ifdef TMS_DEBUG_LOG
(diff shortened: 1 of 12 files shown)


csrc/disk_backend.cpp
```diff
@@ -0,0 +1,95 @@
...
+        CUDA_ERROR_CHECK(cudaMallocHost(&staging_buf_, chunk_bytes_));
+        SIMPLE_CHECK(staging_buf_ != nullptr, "failed to allocate pinned disk staging buffer");
+    }
+}
+
+void DiskBackend::offload(void* gpu_ptr, size_t size, DiskBackupSlot& slot) {
+    ensure_staging_buf_();
+    // Open + preallocate once; reuse the fd and overwrite in place on later pauses.
+    if (slot.fd < 0) {
+        slot.path = dir_ + "/tms_" + std::to_string((long) getpid()) + "_" + std::to_string(counter_++) + ".bin";
```
fzyzcjy: nit: make it a method

torch_memory_saver/entrypoint.py
```diff
@@ -109,17 +125,15 @@ def __init__(self, hook_mode: HookMode = "preload"):
...
     @contextmanager
-    def region(self, tag: str, enable_cpu_backup: bool):
-        # For hook_mode=preload, we need this b/c https://github.com/fzyzcjy/torch_memory_saver/pull/20#issuecomment-3047099047
-        # (For hook_mode=torch we may not need it, but currently our primary usage is hook_mode=preload, thus we do this for simplicity)
-        #
-        # A MemPool is bound to the current device at creation; key by device so a
-        # process using several devices (e.g. multiple TP ranks) doesn't reuse one
-        # device's pool for allocations on another (which bypasses the allocator).
-        key = (tag, enable_cpu_backup, torch.cuda.current_device())
+    def region(self, tag: str, enable_cpu_backup: bool, enable_disk_backup: bool = False):
```
fzyzcjy: maybe this `enable_disk_backup=False` can be required instead of optinoal arg

fzyzcjy (approved):
LGTM w/ nits
and btw it would be great to paste logs showing all (regression) tests pass
also will let my agent have a review (to be posted soon by it), and feel free to ping your agent to read it

csrc/disk_backend.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+// Stream `size` bytes between GPU `gpu_ptr` and `fd`, bouncing through `staging`
+// in `chunk`-byte slices; `to_disk` picks the direction.
+void stream_chunked(int fd, void* gpu_ptr, size_t size, void* staging, size_t chunk, bool to_disk) {
+    size_t offset = 0;
+    while (offset < size) {
+        size_t n = std::min(chunk, size - offset);
+        if (to_disk) {
+            CUDA_ERROR_CHECK(cudaMemcpy(staging, (uint8_t*) gpu_ptr + offset, n, cudaMemcpyDeviceToHost));
+            pwrite_all(fd, staging, n, (off_t) offset);
```
fzyzcjy, line 56: [P1] Buffered `pwrite` does not bound host RAM to the staging chunk. The bytes are also retained in dirty/clean page cache, which is charged to the workload's memory cgroup; with the default `/tmp`, a tmpfs mount stores the entire backup in RAM/swap. Thus the exact workload this mode targets (tens of GB that do not fit in host RAM) can still OOM despite a 256 MiB staging buffer. Please require/validate a real disk-backed directory and control cache growth (for example direct I/O with the required alignment, or writeback plus cache eviction/advice), or narrow the public guarantee to bounded *pinned-buffer* memory rather than bounded host RAM.

csrc/core.cpp
```diff
@@ -3,15 +3,17 @@
...
+TorchMemorySaver::TorchMemorySaver()
+    : disk_backend_(compute_disk_backup_dir_from_env(), compute_disk_chunk_bytes_from_env()) {}
 
 TorchMemorySaver &TorchMemorySaver::instance() {
     static TorchMemorySaver instance;
     return instance;
 }
 
-cudaError_t TorchMemorySaver::malloc(void **ptr, CUdevice device, size_t size, const std::string& tag, const bool enable_cpu_backup) {
+cudaError_t TorchMemorySaver::malloc(void **ptr, CUdevice device, size_t size, const std::string& tag, const bool enable_cpu_backup, const bool enable_disk_backup) {
```
fzyzcjy, line 14: [P1] Enforce CPU/disk mutual exclusion at this native allocation boundary. The Python check is an `assert`, so `TMS_INIT_ENABLE_CPU_BACKUP=1` together with `TMS_INIT_ENABLE_DISK_BACKUP=1`, direct C API use, or `python -O` reaches this function with both flags true. `pause()` then silently takes the CPU branch and allocates the full pinned copy, defeating disk mode and potentially causing the host OOM this feature is meant to avoid; `get_cpu_backup_pointer()` nevertheless returns null because the disk flag is set. A backend enum would make the illegal state unrepresentable, or this function should at least fail fast for every entry point.

torch_memory_saver/entrypoint.py
```diff
@@ -23,10 +23,20 @@ def __init__(self):
...
+        """Context manager for memory saving with optional tag.
+
+        enable_disk_backup spills paused memory to files under disk_backup_dir
+        instead of a pinned CPU buffer; mutually exclusive with enable_cpu_backup.
+        """
         self._ensure_initialized()
-        with self._impl.region(tag=tag, enable_cpu_backup=enable_cpu_backup):
+        assert not (enable_cpu_backup and enable_disk_backup), \
+            "enable_cpu_backup and enable_disk_backup are mutually exclusive"
+        if disk_backup_dir is not None:
```
fzyzcjy: [P1] `disk_backup_dir` is presented as region-scoped, but this mutates one process-global directory and no allocation records its directory. The path is not chosen until the first `pause()`. Consequently, allocating in directory A, then allocating in directory B, then pausing writes both groups to B; nested/concurrent regions have the same problem, and the MemPool key also omits the directory. This can fill the wrong mount or fail after another region removes its directory. Please capture the directory in the allocation/slot (and pool identity where needed), or remove the region parameter and document an immutable process-global configuration.

csrc/disk_backend.cpp
fzyzcjy: [P1] `getpid() + counter` is not globally unique and the following `open(..., O_CREAT)` accepts an existing path. Separate containers commonly have the same PID namespace values while sharing a node-local mount, so two ranks can open the same inode and concurrently overwrite it, then restore mixed model bytes. Stale files after a crash/PID reuse and predictable-path attacks have the same root cause. Please create the file atomically with a random unique name (`mkstemp`, `O_TMPFILE`, or equivalent `O_EXCL`/`O_NOFOLLOW`; also use `O_CLOEXEC`) instead of accepting an existing predictable path.

csrc/disk_backend.cpp
```diff
@@ -0,0 +1,95 @@
...
+        SIMPLE_CHECK(staging_buf_ != nullptr, "failed to allocate pinned disk staging buffer");
+    }
+}
+
+void DiskBackend::offload(void* gpu_ptr, size_t size, DiskBackupSlot& slot) {
+    ensure_staging_buf_();
+    // Open + preallocate once; reuse the fd and overwrite in place on later pauses.
+    if (slot.fd < 0) {
+        slot.path = dir_ + "/tms_" + std::to_string((long) getpid()) + "_" + std::to_string(counter_++) + ".bin";
+        slot.fd = open(slot.path.c_str(), O_RDWR | O_CREAT, 0600);
```
fzyzcjy: [P2] Every disk-backed allocator block keeps this fd open across resume and the entire active lifetime; it is closed only when the underlying allocator block is finally freed. PyTorch MemPool caches those blocks, so the number of live fds can grow with allocator fragmentation rather than live tensors. Once it reaches `RLIMIT_NOFILE`, a later pause fails partway through and `SIMPLE_CHECK` terminates the process. Store the path and reopen/close around each transfer, use a bounded fd cache, or move to a bounded shared-file design.

test/examples/disk_backup.py
```diff
@@ -0,0 +1,62 @@
...
+    try:
+        _run(disk_dir)
+    finally:
+        shutil.rmtree(disk_dir, ignore_errors=True)
+
+
+def _run(disk_dir: str):
+    print("Allocate tensor_with_disk_backup")
+    with torch_memory_saver.region(enable_disk_backup=True, disk_backup_dir=disk_dir):
+        # Larger than the default staging chunk to exercise multi-chunk streaming.
```
fzyzcjy: [P2] This test does not exercise the path the comment claims: 20,000,000 bytes is about 19 MiB, far below the default 256 MiB chunk, so `stream_chunked()` runs only once. The large tensor is then checked at only three bytes, while the full-tensor oracle is another disk-backed tensor rather than an independent CPU copy. Please set the chunk env to a small value before backend initialization and compare a multi-chunk, non-even-tail tensor completely against a CPU oracle; also assert a nonzero expected file count/size so an empty directory cannot satisfy the footprint check.

fzyzcjy (changes requested):
**AI disclosure:** This review was generated and submitted by OpenAI Codex; it is not a human review.

I am requesting changes because the current implementation does not yet provide the advertised bounded-host-memory behavior, and several configuration/resource-lifecycle paths can write to the wrong place, corrupt a backup, or terminate a large job.

The six inline comments below are the blocking findings I retained after reviewing correctness, lifecycle/concurrency, architecture, performance, and tests.

Validation: `git diff --check` and Python byte-compilation passed locally. I could not run the GPU tests in this environment, and GitHub currently reports no checks for this branch.

Zhichenzzz:
Thanks @fzyzcjy! I have addressed the comments and added the test results into the description

fzyzcjy (approved):
LGTM, waiting for miles to pin commit for reproducibility before merging

Zhichenzzz:
already pinned, and yueming approved that miles pr (and we also support the another streaming-offloading stacked on that)

PR #81, opened 2026-07-08
Key MemPool by device so multi-device processes don't share a pool

siju-samuel:
A torch MemPool binds to the device that is current when it is created, and use_mem_pool only redirects allocations on that device. _mem_pools was keyed by (tag, enable_cpu_backup), so a process using several devices (e.g. multiple TP ranks) reused the first device's pool for allocations on another device -- those allocations bypass the custom allocator and are not paused/resumed.

Add the current device to the pool key so each device gets its own pool.

Add a torch-mode multi-device test (multi_device_torch_mode.py) that pins each device at alloc time and asserts both devices free on pause / re-commit on resume; it fails without this fix. Skips when <2 devices are available.

### DIFF
--- a/test/examples/multi_device_torch_mode.py
+++ b/test/examples/multi_device_torch_mode.py
@@ -1,0 +1,71 @@
+"""Multi-device pause/resume in hook_mode="torch".
+
+A torch MemPool binds to the device that is current when it is created, and
+use_mem_pool only redirects allocations on that device. So each pauseable
+allocation must be made with its target device current (the realistic pattern:
+one process pins one device per TP rank).
+
+This protects the fix that keys _mem_pools by device: without it, the pool
+created for device 0 is reused for device 1, the device-1 allocation bypasses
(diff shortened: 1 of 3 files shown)


fzyzcjy (approved):
LGTM, could you plz let agent run the full test suite w/ and w/o fix and paste results here to ensure it works + no regression.

siju-samuel:
master branch - [run_without_fix.txt](https://github.com/user-attachments/files/30092842/run_without_fix.txt)
master branch + my testcase - [run_with_mytestcase_no_fix(failed).txt](https://github.com/user-attachments/files/30092827/run_with_mytestcase_no_fix.failed.txt)
master branch + fix + my_testcase [run_with_fix_mybranch.txt](https://github.com/user-attachments/files/30092829/run_with_fix_mybranch.txt)

@fzyzcjy Thanks for the review. 
Please find the attached test results with and w/o my changes.

PR #82, opened 2026-07-10
Handle CUDA VMM allocation granularity

kaixih:
## Summary

- query the CUDA VMM allocation granularity for the exact `CUmemAllocationProp` used by TMS
- keep the caller-requested logical size separate from the rounded physical VMM size
- use the physical size consistently for create, reserve, map, access, unmap, resume, and address-free
- add a focused CUDA regression test covering an unaligned allocation through pause, resume, and free

## Root cause

TMS currently forwards the original `cudaMalloc` size directly to `cuMemCreate`. CUDA requires that size to be a multiple of the minimum allocation granularity for the selected allocation properties. On the tested GB200 system, the GPUDirect-RDMA-capable VMM property reports a 2 MiB minimum granularity, so any smaller or otherwise unaligned request fails with `CUDA_ERROR_INVALID_VALUE`. The observed 6912-byte request is one representative example; a 128-byte request has the same issue.

This change does not hardcode GB200 or a 2 MiB boundary. It asks the CUDA driver for the applicable granularity at runtime and only rounds requests that need it. Already aligned CUDA allocations are unchanged, and the ROCm path continues to use the requested size directly.

CPU backup and data copies still use the logical size, so padding is never exposed to callers.

## Validation

- Before the fix, the representative 6912-byte test fails at `cuMemCreate` with `CUDA_ERROR_INVALID_VALUE` on GB200.
- After rebuilding TMS from this branch, the same test completes allocation, pause, resume, and free without an external compatibility shim.
- A focused CPU-backup run also verified that data survives pause/resume for the same unaligned allocation.
- Existing focused checks also pass:
  - simple and CPU-backup tests: 4 passed
  - CUDA graph tests: 2 passed
  - configuration and utility tests: 5 passed


fzyzcjy:
Good point, feel free to ping me when ready to merge, and if I do not reply on github feel free to ping on slack

kaixih:
> Good point, feel free to ping me when ready to merge, and if I do not reply on github feel free to ping on slack

Thanks! This is ready now. I verified it on GB200: the regression test fails on the base commit and passes with this change, and the related tests pass as well.

test/test_cuda_vmm_granularity.py
fzyzcjy, line 1: nit: iirc there is a convention to execute tests, would be great to follow that

csrc/core.cpp
```diff
@@ -15,14 +15,15 @@ cudaError_t TorchMemorySaver::malloc(void **ptr, CUdevice device, size_t size, c
     return ROCmHIPImplementation::rocm_malloc(ptr, device, size, tag, enable_cpu_backup, allocation_metadata_, allocator_metadata_mutex_);
 
 #else
+    const size_t allocation_size = CUDAUtils::cu_mem_get_allocation_size(size, device);
```
fzyzcjy: nit: maybe we can rename `size` everywhere into `raw_size`, s.t. we will be more clear there is a separation b/t the two

fzyzcjy (commented):
LGTM and only a bit of nits

test/test_cuda_vmm_granularity.py
kaixih: Done.

csrc/core.cpp
kaixih: Done.

fzyzcjy:
could you please resolve merge conflicts, and I will have a final review as soon as having time

kaixih:
> could you please resolve merge conflicts, and I will have a final review as soon as having time

done. the conflicts are resolved. PTAL.

kaixih:
gentle ping @fzyzcjy

fzyzcjy:
pretty busy recently and will check ASAP

fzyzcjy (changes requested):
This review was performed by an AI coding agent (Codex), with independent correctness, lifecycle, architecture, adversarial, and performance passes. I reviewed commit `9df6230c280b5e7e3b8e366a5caa4ed9759078d8`.

Requesting changes for two P2 issues:

1. **Align the ROCm >= 7 VMM path as well** (`csrc/utils.h:98-101`, exercised through `csrc/core.cpp:18,35-42`). The new ROCm implementation of `cu_mem_get_allocation_size()` returns `raw_size` unchanged, but HIP VMM has the same property-specific granularity requirement: AMD's HIP 7.0 VMM guide says an unaligned `hipMemCreate()` size returns an out-of-memory error and shows querying `hipMemGetAllocationGranularity()` before rounding. Thus a request such as the 6,912-byte regression case still fails on the ROCm >= 7 generic path, while `test/test_examples.py:64-70` explicitly skips ROCm. This runtime behavior already existed in the base, so I am not calling it a new regression; however, this PR adds the cross-platform allocation-size abstraction and explicitly codifies the incorrect identity implementation. Please query and round with the exact `hipMemAllocationProp` used by `hipMemCreate()`, retain `raw_size` for logical copies/bounds, and add corresponding ROCm >= 7 coverage. Reference: https://rocm.docs.amd.com/projects/HIP/en/docs-7.0.0/how-to/hip_runtime_api/memory_management/virtual_memory.html

2. **Do not repeat immutable VMM queries on every allocation** (`csrc/utils.h:128-143`). `cu_mem_get_allocation_size()` calls `cuDeviceGetAttribute()` plus `cuMemGetAllocationGranularity()`, then the immediately following `cu_mem_create()` reconstructs the same property and calls `cuDeviceGetAttribute()` again. Relative to the base, every intercepted allocation now pays one extra device-attribute query and one granularity query in the allocator hot path. These values depend only on the device and fixed allocation property, so please cache a canonical `{property, minimum_granularity}` once per device, thread-safely, and reuse it for both rounding and creation. If repeated querying is intentional, please provide an allocation microbenchmark demonstrating that the added cost is negligible.

The CUDA logical/physical size propagation itself looks consistent: physical VMM operations use `allocation_size`, while CPU backup copies and logical bounds use `raw_size`.

test/examples/cuda_vmm_granularity.py
```diff
@@ -0,0 +1,37 @@
...
+        api, "cudaMalloc", [ctypes.POINTER(ctypes.c_void_p), ctypes.c_size_t]
+    )
+    cuda_free = _bind(api, "cudaFree", [ctypes.c_void_p])
+    cuda_set_device = _bind(api, "cudaSetDevice", [ctypes.c_int])
+    tms_pause = _bind(api, "tms_pause", [ctypes.c_char_p], None)
+    tms_resume = _bind(api, "tms_resume", [ctypes.c_char_p], None)
+
+    assert cuda_set_device(0) == 0
+    ptr = ctypes.c_void_p()
+    assert cuda_malloc(ctypes.byref(ptr), _UNALIGNED_SIZE) == 0
```
fzyzcjy, line 60: nit: maybe we can read/write into that ptr to ensure it is valid

fzyzcjy (approved):
LGTM w/ nits
(after resolving conflicts + add this nit test + paste a result of passing all regression tests I think it is good to merge)

test/examples/cuda_vmm_granularity.py
kaixih: Done. PTAL

kaixih:
Regression results for commit `87ad370` (current PR head):

Environment:

- Miles latest container
- 4x NVIDIA GB200 visible; tests used `CUDA_VISIBLE_DEVICES=0,1`
- PyTorch 2.11.0+cu130
- CUDA 13.0

Focused review coverage:

```text
test_cuda_vmm_granularity
test_disk_backup[preload]
test_disk_backup[torch]

3 passed, 1 warning in 6.40s
```

Full regression suite:

```text
TMS_INIT_ENABLE=0 \
TMS_INIT_ENABLE_CPU_BACKUP=0 \
TMS_INIT_ENABLE_DISK_BACKUP=0 \
CUDA_VISIBLE_DEVICES=0,1 \
python -m pytest -vv -s test

21 passed, 1 warning in 57.13s
```

The warning is the existing `pynvml` deprecation warning.

As a red/green sanity check, latest master (`39093d6`) with the same new
`cuda_vmm_granularity` test fails as expected:

```text
[torch_memory_saver.cpp] CUresult error: 1 (invalid argument)
file=csrc/utils.h func=cu_mem_create line=124
```

The PR head passes that test, including the pointer write/read checks both
before and after pause/resume.

csrc/utils.h
```diff
@@ -94,17 +95,22 @@ namespace CUDAUtils {
             return hipSuccess;
         }
 
-        static void cu_mem_set_access(void *ptr, size_t size, CUdevice device) {
+        static size_t cu_mem_get_allocation_size(size_t raw_size, CUdevice device) {
+            (void)device;
+            return raw_size;
+        }
```
XinyuJiangCMU, line 101: The ROCm path still returns `raw_size` directly here, so the allocation-granularity rounding only applies to CUDA. I'm working on a follow-up to mirror this logic on the HIP path.

XinyuJiangCMU: Follow-up here: #109

PR #84, opened 2026-07-11
Fix ROCm 7.2: free VA in pause() so physical memory is actually released

### LINKED ISSUE #83: ROCm 7.2: pause() frees no physical GPU memory (regression vs ROCm 7.0)
## Summary

On **ROCm 7.2**, `torch_memory_saver.pause(tag)` returns successfully but does **not** free the physical GPU memory back to the driver — `torch.cuda.mem_get_info()` free stays unchanged. The subsequent `resume()` then OOMs on `hipMemCreate` because the memory was never released. The **identical source build works correctly on ROCm 7.0** (frees the memory), so this is a ROCm 7.2 regression in the `cuMemUnmap`/`cuMemRelease` (→ `hipMemUnmap`/`hipMemRelease`) VMM path.

## Minimal repro

```python
import os, torch, torch_memory_saver
tms = torch_memory_saver.torch_memory_saver

def free_gb():
    f, t = torch.cuda.mem_get_info(); return f/2**30, (t-f)/2**30

torch.cuda.init()
print("baseline:   free=%.1f used=%.1f" % free_gb())
with tms.region(tag="kv_cache"):
    x = torch.empty(int(30*2**30/2), dtype=torch.float16, device="cuda")  # ~30GB
torch.cuda.synchronize()
print("after alloc: free=%.1f used=%.1f" % free_gb())
tms.pause("kv_cache"); torch.cuda.synchronize()
print("after PAUSE: free=%.1f used=%.1f  <-- should jump UP" % free_gb())
tms.resume("kv_cache"); torch.cuda.synchronize()
print("after RESUME:free=%.1f used=%.1f" % free_gb())
print("enabled=", tms.enabled)
```

Run in preload mode: `LD_PRELOAD=/path/to/torch_memory_saver_hook_mode_preload.abi3.so python3 repro.py`

## Results (same repro, same source build, same GPU / cluster — only ROCm version differs)

**ROCm 7.0** — HIP `7.0.51831-a3e329ad8`, torch `2.9.0a0+git7bcbafe`:
```
after alloc: free=225.2GB used=30.7GB
after PAUSE: free=255.2GB used=0.7GB   <-- freed ~30GB  ✓ WORKS
after RESUME:free=225.2GB used=30.7GB
```

**ROCm 7.2** — HIP `7.2.26015-fc0010cf6a`, torch `2.9.1+rocm7.2.0`:
```
after alloc: free=225.2GB used=30.7GB
after PAUSE: free=225.2GB used=30.7GB   <-- freed 0GB   ✗ BROKEN
```
`enabled=True` in both; both `torch_memory_saver 0.0.9.post1` built from source and HIP-linked (`ldd` shows `libamdhip64.so.7`, no `libcuda`). CUDA-graph pause not involved (single tagged tensor).

## Environment

- torch_memory_saver 0.0.9.post1, built from source (`pip install .`) with hipcc, preload hook mode.
- GPU: AMD MI30x-class, 256 GB HBM.
- ROCm 7.0 image: `libamdhip64.so.7`, works. ROCm 7.2 image: `libamdhip64.so.7`, broken.

## Impact

Breaks the verl + SGLang colocated RL flow on ROCm 7.2: SGLang's KV-cache pool is never actually released during the training phase (`release_memory_occupation` calls `pause("kv_cache")` which frees nothing), so `resume_memory_occupation` later fails with `hipMemCreate hipErrorOutOfMemory` (`csrc/core.cpp func=resume`). Works fine on ROCm 7.0.

## Ask

Does the ROCm 7.0+ single-allocation VMM path (`cuMemUnmap`+`cuMemRelease`) need adjustment for ROCm 7.2 (e.g. `hipMemRelease` semantics / an extra `hipMemAddressFree` or sync), or is a ROCm 7.2 driver bug the root cause? Happy to run additional diagnostics on ROCm 7.2 hardware.

ChangyiYang:
Fixes #83.

## Problem

On **ROCm 7.2**, `pause()` returns successfully but does **not** free the physical GPU memory — `torch.cuda.mem_get_info()` is unchanged. A later `resume()` then OOMs on `cuMemCreate` (`csrc/core.cpp func=resume`) once the paused allocation is large enough. The **same source build works on ROCm 7.0**, so it's a ROCm 7.2 behavior change in the VMM path: `cuMemUnmap` + `cuMemRelease` no longer return the physical pages while the virtual address range stays reserved.

## Fix

In `pause()` (non-`TMS_ROCM_LEGACY_CHUNKED` path), after `cuMemUnmap` + `cuMemRelease`, also `cuMemAddressFree` the range so the physical memory is actually returned, then **immediately re-reserve the same address** as a placeholder so the allocation pointer stays valid for `resume()`. `resume()` is unchanged (it maps a fresh handle into the already-reserved VA).

This is safe on ROCm 7.0 as well (the VA is simply freed and re-reserved in place), and CUDA behavior is unchanged for the common case since the re-reserved address is verified to match.

## Verification (ROCm 7.2, HIP 7.2.26015, MI30x, 256 GB)

Minimal repro (allocate 30 GB inside a `region`, pause, resume):

| stage | before fix | after fix |
|---|---|---|
| after alloc | free 225.2 GB | free 225.2 GB |
| after `pause()` | free 225.2 GB (freed 0) ❌ | free **255.2 GB** (freed ~30) ✅ |
| after `resume()` | — | free 225.2 GB (re-mapped, same VA) ✅ |

Also validated end-to-end in a colocated verl + SGLang RL run at `gpu_memory_utilization=0.8` on ROCm 7.2: SGLang's KV-cache pool is now actually released during training (~175 GB freed at `release_memory_occupation`), `resume_memory_occupation` no longer OOMs, and training completes multiple steps — a configuration that previously always OOM'd. No regression observed on ROCm 7.0.

### DIFF
--- a/csrc/core.cpp
+++ b/csrc/core.cpp
@@ -131,10 +131,14 @@ void TorchMemorySaver::pause(const std::string& tag) {
 
         CURESULT_CHECK(cuMemUnmap((CUdeviceptr) ptr, metadata.size));
         CURESULT_CHECK(cuMemRelease(metadata.allocHandle));
+#if defined(USE_ROCM)
         // On ROCm 7.2, cuMemUnmap + cuMemRelease do not return the physical pages to
         // the driver while the virtual address range stays reserved (they do on ROCm
-        // 7.0). Free the VA to actually release the memory, then immediately re-reserve
-        // the SAME address as a placeholder so the pointer stays valid for resume().
+        // 7.0 and on CUDA). Free the VA to actually release the memory, then immediately
+        // re-reserve the SAME address as a placeholder so the pointer stays valid for
+        // resume(). ROCm-only: CUDA already frees the physical on release, so it does
+        // not need this and must not be exposed to the (small) free->re-reserve window
+        // where another thread could grab the address.
         CURESULT_CHECK(cuMemAddressFree((CUdeviceptr) ptr, metadata.size));
         {
             CUdeviceptr reserved = 0;
@@ -147,6 +151,7 @@ void TorchMemorySaver::pause(const std::string& tag) {
                 exit(1);
             }
         }
+#endif
 
         metadata.state = AllocationState::PAUSED;
 
--- a/csrc/core.cpp
+++ b/csrc/core.cpp
@@ -131,6 +131,22 @@ void TorchMemorySaver::pause(const std::string& tag) {
 
         CURESULT_CHECK(cuMemUnmap((CUdeviceptr) ptr, metadata.size));
         CURESULT_CHECK(cuMemRelease(metadata.allocHandle));
+        // On ROCm 7.2, cuMemUnmap + cuMemRelease do not return the physical pages to
+        // the driver while the virtual address range stays reserved (they do on ROCm
+        // 7.0). Free the VA to actually release the memory, then immediately re-reserve
+        // the SAME address as a placeholder so the pointer stays valid for resume().
+        CURESULT_CHECK(cuMemAddressFree((CUdeviceptr) ptr, metadata.size));
+        {
+            CUdeviceptr reserved = 0;
+            CURESULT_CHECK(cuMemAddressReserve(&reserved, metadata.size, 0, (CUdeviceptr) ptr, 0));
+            if ((void*) reserved != ptr) {
+                std::cerr << "[torch_memory_saver.cpp] pause() could not re-reserve the same VA:"
+                          << " want=" << ptr << " got=" << (void*) reserved
+                          << " file=" << __FILE__ << " func=" << __func__ << " line=" << __LINE__
+                          << std::endl;
+                exit(1);
+            }
+        }
 
         metadata.state = AllocationState::PAUSED;
 


ChangyiYang:
Thanks for the review — good call on the race exposure.

**Restricted to ROCm** (commit 27bd085): the `cuMemAddressFree` + re-reserve block is now under `#if defined(USE_ROCM)`. CUDA already returns the physical memory on `cuMemRelease`, so it keeps the original `unmap`+`release`-only path and is not exposed to the free→re-reserve window. Re-verified on ROCm 7.2 (HIP 7.2.26015) after gating: `pause()` still frees (30 GB → returned) and `resume()` re-maps at the same address.

**On the race:** it's inherent to this approach — CUDA/HIP VMM has no atomic "free + re-reserve at the same address", so any solution that frees the VA to reclaim physical has this window. Two things bound it in practice: (1) it's now ROCm-only, and (2) in the intended use (verl/SGLang `release_memory_occupation`) `pause()` runs while the server is idle (`is_fully_idle()` is asserted), so there is no concurrent generation issuing allocations at that moment. The `exit(1)` is a fail-loud guard rather than silent corruption. If you'd prefer a fully race-free design I'm happy to explore reserving one large VA arena up front and sub-mapping within it, but that's a bigger change — let me know your preference.

**Log prefix:** I kept `[torch_memory_saver.cpp]` to match the existing `std::cerr` messages in this file (they all use that prefix despite the file being `core.cpp`). Happy to switch them all to `[core.cpp]` in a follow-up if you'd like consistency the other way.

csrc/core.cpp
fzyzcjy, line 1: the fix looks a bit hacky, cc @yushengsu-thu do you want to have a look at this - does this look good to you e.g. it is common to do this in the AMD world

fzyzcjy, line 1: nit: could you please ensure the test suite fails on rocm 7.2 w/o the fix (and pass after the fix), s.t. it will protect in the future

fzyzcjy (commented):
LGTM w/ nits

ChangyiYang:
@fzyzcjy thanks for looking at it. After testing, this hacky solution did fail on some edge cases. We rollback to rocm7.0 for now while AMD folks will take a look at how to property support it on rocm7.2.

fzyzcjy:
I see... feel free to ping me when needed

PR #86, opened 2026-07-23
Retain pinned CPU backups across resume cycles

williamjameshandley:
## Summary

Add an opt-in CUDA policy that retains pinned CPU backups across resume cycles:

- `torch_memory_saver.retain_cpu_backup = True`
- `TMS_RETAIN_CPU_BACKUP=1` at process startup

A later pause refreshes the retained storage in place. The existing backup query exposes the retained snapshot after resume, and CUDA process cleanup releases any remaining backups after memory pools are destroyed. ROCm remains unchanged and rejects enabling the policy.

This avoids repeated pinned-host allocation for large, interruptible model-weight workloads. It deliberately trades pinned RAM equal to backed-up allocations for faster repeated resume.

## Validation

Against current `master` (`6d5bce4`):

- CUDA 13 extensions build successfully on Python 3.14
- the CPU-backup scenario passes with both `torch` and `preload` hook modes on H100 hardware
- environment initialization passes in fresh subprocesses for both hook modes
- restoring unconditional `cudaFreeHost` on resume deterministically fails the retained active-backup assertion
- Python sources compile cleanly and `git diff --check` passes

Production validation on an 8x H100 model server:

- median weight resume: 4.846 s before, 0.623 s with retention
- five release/resume cycles completed with inference canaries and no service restart
- separate opt-in instrumentation verified 136,004,501,504 restored bytes byte-for-byte; that instrumentation is intentionally excluded from this PR

### DIFF
--- a/csrc/core.cpp
+++ b/csrc/core.cpp
@@ -283,18 +283,6 @@ cudaError_t TorchMemorySaver::resume(const std::string& tag) {
 #endif
 }
 
-void TorchMemorySaver::release_cpu_backups() {
-#ifdef USE_CUDA
-    const std::lock_guard<std::mutex> lock(allocator_metadata_mutex_);
-    for (auto& [ptr, metadata] : allocation_metadata_) {
-        if (metadata.cpu_backup != nullptr) {
-            CUDA_ERROR_CHECK(cudaFreeHost(metadata.cpu_backup));
-            metadata.cpu_backup = nullptr;
-        }
-    }
-#endif
-}
-
 uint8_t* TorchMemorySaver::get_cpu_backup_pointer(const uint8_t* query_gpu_ptr, uint64_t query_size) {
     const std::lock_guard <std::mutex> lock(allocator_metadata_mutex_);
 
--- a/csrc/core.cpp
+++ b/csrc/core.cpp
@@ -5,9 +5,7 @@
 
 TorchMemorySaver::TorchMemorySaver() {
 #ifdef USE_CUDA
-    const char* retain_cpu_backup = std::getenv("TMS_RETAIN_CPU_BACKUP");
-    retain_cpu_backup_.store(
-        retain_cpu_backup != nullptr && std::string(retain_cpu_backup) == "1");
+    retain_cpu_backup_.store(get_bool_env_var("TMS_RETAIN_CPU_BACKUP"));
 #endif
 }
 
--- a/csrc/core.cpp
+++ b/csrc/core.cpp
@@ -236,7 +236,9 @@ uint8_t* TorchMemorySaver::get_cpu_backup_pointer(const uint8_t* query_gpu_ptr, 
         if ((ptr <= query_gpu_ptr) && (query_gpu_ptr + query_size <= ptr + total_size)) {
             const size_t offset = query_gpu_ptr - ptr;
             if (metadata.state == AllocationState::ACTIVE) {
-                return nullptr;
+                return metadata.cpu_backup == nullptr
+                    ? nullptr
+                    : (uint8_t*) metadata.cpu_backup + offset;
             } else {
                 SIMPLE_CHECK(nullptr != metadata.cpu_backup,
                     "get_cpu_backup_pointer: found paused allocation but cpu_backup does not exist, do you forget to enable cpu backup");
--- a/csrc/core.cpp
+++ b/csrc/core.cpp
@@ -3,7 +3,13 @@
 #include "macro.h"
 #include "api_forwarder.h"
 
-TorchMemorySaver::TorchMemorySaver() {}
+TorchMemorySaver::TorchMemorySaver() {
+#ifdef USE_CUDA
+    const char* retain_cpu_backup = std::getenv("TMS_RETAIN_CPU_BACKUP");
+    retain_cpu_backup_.store(
+        retain_cpu_backup != nullptr && std::string(retain_cpu_backup) == "1");
+#endif
+}
 
 TorchMemorySaver &TorchMemorySaver::instance() {
     static TorchMemorySaver instance;
@@ -151,6 +157,7 @@ void TorchMemorySaver::resume(const std::string& tag) {
 
 #else
     const std::lock_guard <std::mutex> lock(allocator_metadata_mutex_);
+    const bool retain_cpu_backup = retain_cpu_backup_.load();
 
     for (auto it = allocation_metadata_.begin(); it != allocation_metadata_.end(); ++it) {
         void *ptr = it->first;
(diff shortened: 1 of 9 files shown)


torch_memory_saver/entrypoint.py
fzyzcjy, line 1: nit: given retain_cpu_backup looks like sth similar to other flags such as enable-cpu-backup, enable-disk-backup, etc, maybe we can make the configuration code mimic them

test/examples/cpu_backup.py
```diff
@@ -38,6 +39,40 @@ def run(hook_mode: str):
...
+    # Tags remain independent, and retained host storage is reused at the same
+    # address while its bytes are refreshed. Exercise every visible device.
+    for device in range(torch.cuda.device_count()):
+        with torch.cuda.device(device):
+            with torch_memory_saver.region(tag=f"retained_{device}", enable_cpu_backup=True):
+                selected = torch.full((1024,), 40 + device, dtype=torch.uint8, device=device)
+            with torch_memory_saver.region(tag=f"other_{device}", enable_cpu_backup=True):
+                other = torch.full((1024,), 80 + device, dtype=torch.uint8, device=device)
+
+            torch_memory_saver.retain_cpu_backup = True
```
fzyzcjy: [P1] This generic CPU-backup scenario also runs on ROCm, where assigning `True` reaches `SIMPLE_CHECK(!retain_cpu_backup, ...)` and terminates the process with `exit(1)`. Both ROCm `test_cpu_backup` variants therefore abort; in torch hook mode the parent then blocks indefinitely at `output_queue.get()` because the child exits before it can enqueue a result. The new env-policy test has the complementary failure: it expects `True`, while the ROCm getter always returns `False`. Please keep the generic ROCm scenario unchanged and move/skip all retention assertions in CUDA-only tests.

csrc/core.cpp
```diff
@@ -217,7 +234,9 @@ uint8_t* TorchMemorySaver::get_cpu_backup_pointer(const uint8_t* query_gpu_ptr,
         if ((ptr <= query_gpu_ptr) && (query_gpu_ptr + query_size <= ptr + total_size)) {
             const size_t offset = query_gpu_ptr - ptr;
             if (metadata.state == AllocationState::ACTIVE) {
-                return nullptr;
+                return metadata.cpu_backup == nullptr
+                    ? nullptr
+                    : (uint8_t*) metadata.cpu_backup + offset;
```
fzyzcjy, line 308: [P1] This ACTIVE-state return is compiled for ROCm too, even though the new policy is advertised as CUDA-only. The ROCm 6.x resume path restores from `metadata.cpu_backup` but does not free or null it, so after this change `get_cpu_backup()` starts returning a retained/stale snapshot for ACTIVE ROCm allocations while `retain_cpu_backup` still reports `False` and enabling it is rejected. ROCm 7.x behaves differently because it follows this file's non-retaining resume path. Please gate the new ACTIVE visibility to CUDA or first normalize the ROCm backup lifetime.

torch_memory_saver/entrypoint.py
```diff
@@ -83,6 +83,16 @@ def memory_margin_bytes(self, value: int):
...
 
+    @property
+    def retain_cpu_backup(self) -> bool:
+        self._ensure_initialized()
+        return self._impl._binary_wrapper.cdll.tms_get_retain_cpu_backup()
+
+    @retain_cpu_backup.setter
+    def retain_cpu_backup(self, value: bool):
+        self._ensure_initialized()
+        self._impl._binary_wrapper.cdll.tms_set_retain_cpu_backup(value)
```
fzyzcjy, line 113: [P2] Treating this as an immediate runtime call makes configuration order observable in a surprising way: `torch_memory_saver.retain_cpu_backup = True; torch_memory_saver.hook_mode = "torch"` initializes the default preload implementation on the first statement, deletes `_impl_ctor_kwargs`, and makes the second statement fail. Merely reading the new getter has the same side effect. Please put this policy into the pending initialization config (alongside `hook_mode`) and apply it when the implementation is created, so inspecting or setting one config value does not silently choose the allocator backend.

torch_memory_saver/entrypoint.py
```diff
@@ -107,6 +117,15 @@ def __init__(self, hook_mode: HookMode = "preload"):
             # destruction order fiasco"). By clearing _mem_pools in an atexit handler, we ensure MemPool 
             # destruction (and thus HIP API calls) happens while the HIP/HSA runtime is still fully alive.
             atexit.register(self._mem_pools.clear)
+        else:
+            # Destroy pools while CUDA and the allocator configuration are
+            # alive, then release any backups not owned by a Python MemPool.
+            atexit.register(self._cleanup_cuda_at_exit)
+
+    def _cleanup_cuda_at_exit(self):
+        self._binary_wrapper.cdll.tms_set_interesting_region(True)
```
fzyzcjy: [P2] This permanently changes the current thread's allocator policy. Python runs `atexit` handlers in LIFO order, so an application handler registered before lazy TMS initialization will run after this cleanup; any CUDA work in that later handler can now be intercepted even if the original `interesting_region` value was false, and it happens after the one-shot pool/backup cleanup has already run. Please save the original value and restore it in `finally`, or avoid mutating allocator policy as part of shutdown.

fzyzcjy (changes requested):
(see the comment below about this ai review)

This review was performed by an AI coding agent (Codex), not a human reviewer. I found four blocking regressions/design issues; details are inline. I reviewed head `098b0e9` against its merge base and also checked integration with current `master` (`39093d6`). The PR is currently conflicting with `master`, but the comments below are semantic issues in the PR itself rather than mechanical merge conflicts. I did not find an additional hot-path performance regression beyond the documented opt-in pinned-RAM tradeoff.

test/examples/cpu_backup.py
fzyzcjy: can ignore rocm

fzyzcjy:
(I will make a final manual review after the human suggestion above is resolved; the agent's auto review is for reference, not a strict requirement)

fzyzcjy (approved):

fzyzcjy:
Validated PR head `ed1f26d7e3e1af15f7f514a61e8c99685ae9b204` on a clean isolated CUDA environment.

Environment:
- Host: `tom-workstation`
- Container image: `m.daocloud.io/docker.io/pytorch/pytorch:2.6.0-cuda12.4-cudnn9-devel`
- Image ID: `sha256:7d57e307bd9c152da255651dc3458beaecf35fcdf4752fbbc14217b9be0307aa`
- GPU: `NVIDIA GeForce RTX 4090 D` (GPU 0 only)
- Driver: `550.163.01`
- Python: `3.11.11`
- PyTorch: `2.6.0+cu124`
- CUDA runtime: `12.4`
- CUDA toolkit: `12.4.131`

Build:
```bash
TMS_CUDA_MAJOR=12 python setup.py build_ext --inplace
```
Result: both preload and torch hook-mode extensions built successfully.

Targeted retained-backup regression:
```bash
CUDA_VISIBLE_DEVICES=0 python -m pytest test/test_examples.py::test_cpu_backup -vv -ra -s
```
Result: `2 passed in 2.54s`.

This covers both hook modes, resume-time policy changes, non-eager disable followed by release on a non-retaining resume, full non-uniform backup/restore contents, retained host pointer reuse, and pinned-host release when ACTIVE and PAUSED allocations are torn down.

Full regression suite:
```bash
CUDA_VISIBLE_DEVICES=0 python -m pytest test -vv -ra
```
Result: `22 passed, 9 skipped in 44.10s`.

All skips were expected for this single-GPU CUDA environment: 3 multi-device tests require at least two visible GPUs, and 6 tests are XPU-only. There were no failures or unexpected skips.

PR #87, opened 2026-07-27
Propagate out-of-memory from resume() instead of calling exit()

yueming-yuan:
## Problem

`CUDAUtils::cu_mem_create` deliberately treats OOM as recoverable, and says so:

```cpp
CUresult ret = cuMemCreate(alloc_handle, size, &prop, 0);
if (ret == CUDA_ERROR_OUT_OF_MEMORY) {
    std::cerr << "[torch_memory_saver.cpp] cuMemCreate CUDA_ERROR_OUT_OF_MEMORY (may not be an issue e.g. torch allocator will free cache and retry)" << std::endl;
    return cudaErrorMemoryAllocation;
}
```

It has two call sites in `csrc/core.cpp`, and they disagree:

```cpp
// line 40, allocation path — honours the contract
cudaError_t ret = CUDAUtils::cu_mem_create(&allocHandle, size, device);
if (ret != cudaSuccess) {
    return ret;
}

// line 185, resume path — turns it back into process death
CUDA_ERROR_CHECK(CUDAUtils::cu_mem_create(&newAllocHandle, metadata.size, metadata.device));
```

`CUDA_ERROR_CHECK` calls `exit(1)`. `exit()` then runs static destructors, and torch's `CudaIPCGlobalEntities` destructor touches a CUDA context that is no longer usable, so the process dies with `SIGABRT` and **no Python traceback**:

```
Fatal Python error: Aborted

Current thread ...:
  File ".../torch_memory_saver/entrypoint.py", line 169 in resume
  ...
```

We hit this repeatedly on a 64-rank colocated RL job that pauses the inference engine's weights during training and resumes them afterwards. Under memory pressure a single rank would vanish with no diagnosis and take the whole job with it. It took a while to identify because the symptom — no traceback, no OOM in the kernel log, random timing — looks nothing like an allocation failure.

## Change

- `TorchMemorySaver::resume` returns `cudaError_t` and propagates the `cu_mem_create` failure.
- `tms_resume` forwards it.
- The Python wrapper declares `restype` and raises `torch.cuda.OutOfMemoryError` on `cudaErrorMemoryAllocation` (and `RuntimeError` otherwise), so a caller can `torch.cuda.empty_cache()` and retry — which is exactly what the allocation path already relies on.

One more change is needed for a retry to actually work. `resume()` loops over allocations, so a resume that returns partway leaves the earlier ones `ACTIVE`; the not-`PAUSED` branch used to `exit(1)` on exactly those, so the retry would abort. It now skips them, making `resume()` idempotent.

## Compatibility

The C ABI change is additive — existing callers that ignore the return value (sglang, slime, and others) keep working unchanged.

## Verified

Builds clean for CUDA 13 on aarch64, both hook modes:

```
torch_memory_saver_hook_mode_preload_cu13.abi3.so
torch_memory_saver_hook_mode_torch_cu13.abi3.so
```

Only pre-existing `-Wunused-function` warnings. I have not written a test that forces the OOM path — reproducing it needs a GPU deliberately driven to the edge of capacity — so a review of the error-path logic would be welcome.

## Out of scope

The ROCm legacy chunked path (`ROCmHIPImplementation::rocm_resume`) has its own copy of the not-`PAUSED` `exit(1)` and still returns void; `resume()` returns `cudaSuccess` after calling it. Happy to extend this to that path if you would prefer it in one PR.

### DIFF
--- a/csrc/core.cpp
+++ b/csrc/core.cpp
@@ -158,9 +158,10 @@ void TorchMemorySaver::pause(const std::string& tag) {
 #endif
 }
 
-void TorchMemorySaver::resume(const std::string& tag) {
+cudaError_t TorchMemorySaver::resume(const std::string& tag) {
 #if TMS_ROCM_LEGACY_CHUNKED
     ROCmHIPImplementation::rocm_resume(tag, allocation_metadata_, allocator_metadata_mutex_);
+    return cudaSuccess;
 
 #else
     const std::lock_guard <std::mutex> lock(allocator_metadata_mutex_);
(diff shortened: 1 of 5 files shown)


csrc/core.cpp
fzyzcjy, line 1: nit: for the no log issue, wondering why std::cerr print does not work. shall we somehow ensure text is flushed, and if it is issue of ray log forwarding or sth, shall we log to a provided file or kernel log etc

csrc/core.cpp
```diff
@@ -174,15 +175,20 @@ void TorchMemorySaver::resume(const std::string& tag) {
...
 
         if (metadata.state != AllocationState::PAUSED) {
-            std::cerr << "[torch_memory_saver.cpp] Cannot resume allocation that is not paused. "
-                      << " tag=" << metadata.tag << " ptr=" << std::to_string((uintptr_t)ptr)
-                      << " file=" << __FILE__ << " func=" << __func__ << " line=" << __LINE__
-                      << std::endl;
-            exit(1);
+            // Already active. Skipping rather than aborting keeps resume()
+            // idempotent, which is what lets a caller retry after resume()
+            // returned an out-of-memory error partway through the map.
```
fzyzcjy, line 180: would be great to add tests like

1. vanilla case (protect the issue reported)
2. for such "lets a caller retry after resume() returned an out-of-memory error", maybe also add a specific test for it

PR #94, opened 2026-08-05
Unset change_env keys that were previously absent

### LINKED ISSUE #91: change_env restores unset keys as empty strings
## Problem

`torch_memory_saver.utils.change_env` restores previously absent environment variables as `\"\"` instead of unsetting them. C++ `get_bool_env_var` rejects empty values and aborts, so a later call that reads `TMS_INIT_*` can fail after an earlier `change_env` on an unset key.

This shows up when running the test suite in order (e.g. a test that uses `change_env(\"TMS_INIT_ENABLE\", \"1\")` followed by another test that relies on the key remaining unset).

## Minimal reproduction

Repro commit on branch [`repro/change-env-empty-restore`](https://github.com/modal-projects/torch_memory_saver/tree/repro/change-env-empty-restore):

```bash
python test/examples/change_env_empty_restore.py
```

Expected: key is absent after the context manager.
Actual: key is present with value `\"\"`.

## Environment

- Any platform; Python-only until a native bool env read sees the empty value

## Expected behavior / ask

If the key was absent before `change_env`, leave it absent on exit (or otherwise avoid writing an empty string that native bool parsing rejects).

botirkhaltaev:
## Summary

Closes #91.

- `change_env` now tracks whether the key was absent (`None`) vs present
- On exit, absent keys are popped instead of restored as `""` (which breaks C++ `get_bool_env_var`)
- Previously set values are still restored

## Test plan

- [x] `python -m pytest test/test_utils.py` (includes absent-key and restore-value cases)

### DIFF
--- a/torch_memory_saver/utils.py
+++ b/torch_memory_saver/utils.py
@@ -97,8 +97,6 @@ def change_env(key: str, value: str):
         yield
     finally:
         assert os.environ[key] == value
-        # Restoring an absent key as "" would break consumers that reject empty
-        # values (e.g. C++ get_bool_env_var aborts on them).
         if old_value is None:
             os.environ.pop(key, None)
         else:
--- a/torch_memory_saver/utils.py
+++ b/torch_memory_saver/utils.py
@@ -90,12 +90,17 @@ def get_binary_path_from_package(stem: str):
 # private utils, not to be used by end users
 @contextmanager
 def change_env(key: str, value: str):
-    old_value = os.environ.get(key, "")
+    old_value = os.environ.get(key)
     os.environ[key] = value
     logger.debug(f"change_env set key={key} value={value}")
     try:
         yield
     finally:
         assert os.environ[key] == value
-        os.environ[key] = old_value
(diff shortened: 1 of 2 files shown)


torch_memory_saver/utils.py
```diff
@@ -90,12 +90,17 @@ def get_binary_path_from_package(stem: str):
...
+    old_value = os.environ.get(key)
     os.environ[key] = value
     logger.debug(f"change_env set key={key} value={value}")
     try:
         yield
     finally:
         assert os.environ[key] == value
-        os.environ[key] = old_value
+        # Restoring an absent key as "" would break consumers that reject empty
+        # values (e.g. C++ get_bool_env_var aborts on them).
```
fzyzcjy: nit: rm this comment

fzyzcjy (approved):
LGTM w/ nits

torch_memory_saver/utils.py
botirkhaltaev: Removed.

botirkhaltaev:
Nit addressed in f6e47b4 (comment removed).

Issue #88, opened 2026-07-31
NCCL_GRAPH_REGISTER=1 can hang after a same-VA TMS resume


kaixih:
## Problem

torch-memory-saver preserves the CUDA virtual address across `pause()` and
`resume()`, but `resume()` maps a new physical allocation at that address.

With the default `NCCL_GRAPH_REGISTER=1`, the relevant lifecycle is:

```text
CUDA graph capture:
  buffer = (VA V, backing P0)
  MR0 = register(V, P0)

TMS resume:
  buffer = (VA V, backing P1)

CUDA graph replay:
  NCCL may find V in its graph-registration cache and reuse MR0
  MR0 was created for P0, while V now maps to P1
  the retained graph can hang
```

We reproduced this hang after TMS resume in a two-node GB200 NCCL workload.

## Current workaround

Set the full NCCL environment variable before CUDA graph capture:

```bash
export NCCL_GRAPH_REGISTER=0
```

With graph registration disabled, the lifecycle becomes:

```text
CUDA graph capture:
  buffer = (VA V, backing P0)
  do not register(V, P0)
  no graph-owned MR0 exists

TMS resume:
  buffer = (VA V, backing P1)

CUDA graph replay:
  the CUDA kernel still accesses VA V
  V now resolves to P1
  there is no stale graph-owned MR for P0 to reuse
```

This setting disables persistent CUDA-graph registration of NCCL user
send/receive buffers. It does **not** disable CUDA graphs, NCCL, or GPUDirect
RDMA registrations for NCCL's own internal transport buffers.

With `NCCL_GRAPH_REGISTER=0`, two consecutive two-rollout/two-update runs
completed successfully while replaying the retained CUDA graphs after every
TMS resume.

## Question for TMS

Should TMS document that backing-bound external registrations must be retired
before `pause()`, or provide `before_unmap` / `after_remap` hooks so consumers
can deregister the old backing and register the new one?

## Related changes

- torch-memory-saver
  [#82](https://github.com/fzyzcjy/torch_memory_saver/pull/82): the pinned CUDA
  VMM implementation used by the reproducer
- Miles [#1726](https://github.com/radixark/miles/pull/1726): the two-node GB200
  HybridEP workflow
- Proposed Miles script change
  [`98528fc`](https://github.com/kaixih/miles/commit/98528fc65b044fa203f094df1d673eb1aacb9b09):
  put `NCCL_GRAPH_REGISTER=0` in the Ray runtime environment
[...]

PR #107, opened 2026-08-19
Document NCCL graph registration caveat

kaixih:
## Summary

- document the NCCL graph-registration risk when a retained CUDA graph uses TMS-managed buffers across `pause()` / `resume()`
- describe both mitigations discussed in #88: keep buffers passed to NCCL outside TMS-managed regions, or set `NCCL_GRAPH_REGISTER=0`
- clarify that the environment variable leaves NCCL and CUDA graphs enabled while potentially giving up the associated zero-copy optimization

## Why

TMS preserves a buffer's virtual address across `pause()` / `resume()` while replacing its physical backing. A retained NCCL CUDA-graph user-buffer registration may therefore become stale and cause graph replay to hang.

Addresses #88.

## Validation

- `git diff --check origin/master...HEAD`
- documentation-only change; no runtime tests run

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,20 @@ Not only does torch_memory_saver make tensors compatible with CUDA graph, but we
 
 API: Change `torch.cuda.graph(...)` to `torch_memory_saver.cuda_graph(...)`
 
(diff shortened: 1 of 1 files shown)


kaixih:
@fzyzcjy for review.

PR #108, opened 2026-08-27
Add $ORIGIN-relative RPATH to CUDA hooks so LD_PRELOAD resolves pip libcudart

rishabhsinha17:
## Summary

Bake an `$ORIGIN`-relative RPATH into the CUDA hook `.so` files so the LD_PRELOAD-ed preload hook resolves the pip-installed `libcudart.so.<major>` without any `LD_LIBRARY_PATH` help. Fixes the startup crash reported in sgl-project/sglang#36533 at the root.

## Root cause

`configure_subprocess()` injects the preload hook into child processes via `LD_PRELOAD`. The dynamic loader resolves the preloaded object's `NEEDED libcudart.so.13` before Python or torch load, so torch's RPATH-carrying libraries cannot supply it. The shipped hook has `NEEDED libcuda.so.1` and `NEEDED libcudart.so.13` but no RPATH/RUNPATH, and pip environments keep the CUDA runtime in `site-packages/nvidia/cu13/lib` (cu12: `nvidia/cuda_runtime/lib`), which is not on the default search path. An unresolvable `NEEDED` of a preloaded object is a fatal loader error, so the child exits 127 with `libcudart.so.13: cannot open shared object file`.

Verified against the shipped 0.0.9.post1 wheel in a torch 2.13.0+cu130 venv (Python 3.11.10, NVIDIA L40S):

```text
readelf -d site-packages/torch_memory_saver_hook_mode_preload_cu13.abi3.so
 0x0000000000000001 (NEEDED)             Shared library: [libcuda.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libcudart.so.13]
 # ... no RPATH/RUNPATH entry
```

and reproduced end-to-end via sglang `--enable-memory-saver` (Qwen2.5-0.5B-Instruct, 5-run loops, `LD_LIBRARY_PATH`/`ldconfig` asserted unable to resolve `libcudart.so.13`): 5/5 scheduler startup crashes with exit code 127, and 5/5 clean startups once the wheel runtime directory is made resolvable to the child loader (sgl-project/sglang#36643).

## Changes

- `setup.py`: CUDA extension builds add `extra_link_args` rpath entries relative to the install location of the hook:
  - `$ORIGIN/nvidia/cu<major>/lib` and `$ORIGIN/../nvidia/cu<major>/lib` for `TMS_CUDA_MAJOR >= 13` (consolidated cu13 wheel layout)
  - `$ORIGIN/nvidia/cuda_runtime/lib` and `$ORIGIN/../nvidia/cuda_runtime/lib` for cu12 (per-component wheel layout)

  The two prefixes cover both lookup locations `utils.get_binary_path_from_package()` supports (site-packages root and the package directory). Nonexistent rpath entries are ignored by the loader, so system or conda CUDA installs are unaffected. HIP and XPU builds are untouched.
- `test/test_setup_rpath.py`: regression tests asserting the cu12 and cu13 extension configs carry the expected rpath link args and HIP stays clean.

Downstream note: sglang is adding a complementary adapter-side fix that prepends the discovered runtime directory to the child's `LD_LIBRARY_PATH` (sgl-project/sglang#36643), which covers wheels already released. This change makes newly built wheels correct on their own.

## Validation

Capture test, verified both ways (fails without the fix, passes with it):

```text
# with master's setup.py + this branch's test
pytest test/test_setup_rpath.py  ->  2 failed, 1 passed

# with this branch
pytest test/test_setup_rpath.py  ->  3 passed
```

Honest scope note: I did not rebuild a cu13 wheel to `readelf` the produced RUNPATH end-to-end (the validation GPU box runs driver 570, so its builds go through a compat shim I did not want in the evidence chain). The link-args capture test pins the exact flags handed to the compiler, and the loader mechanism itself (unresolvable `NEEDED` on a preloaded object is fatal; resolvable runtime dir fixes it) is demonstrated 5/5-both-ways in the sglang validation above.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -124,6 +124,7 @@ def _create_ext_modules(platform):
 
     # Common compile arguments
     extra_compile_args = ['-std=c++17', '-O3']
+    extra_link_args = []
     
     # Platform-specific configurations
     platform_home = Path(_find_platform_home(platform))
@@ -162,6 +163,13 @@ def _create_ext_modules(platform):
                 "(use `make build-wheel-multi-cuda` or scripts/build_multi_cuda.sh)."
             )
         name_suffix = f"_cu{cuda_major}"
+        # RUNPATH (not RPATH) so the LD_PRELOAD-ed hook finds the pip CUDA runtime but LD_LIBRARY_PATH still wins
+        runtime_rel = f"nvidia/cu{cuda_major}/lib" if int(cuda_major) >= 13 else "nvidia/cuda_runtime/lib"
+        extra_link_args = [
+            "-Wl,--enable-new-dtags",
+            f"-Wl,-rpath,$ORIGIN/{runtime_rel}",
+            f"-Wl,-rpath,$ORIGIN/../{runtime_rel}",
+        ]
     else:
         name_suffix = ""
 
@@ -190,6 +198,7 @@ def _create_ext_modules(platform):
             ],
             py_limited_api=True,
             extra_compile_args=extra_compile_args,
+            extra_link_args=extra_link_args,
         )
         for name, extra_macros in hook_variants
     ]
--- a/test/test_setup_rpath.py
+++ b/test/test_setup_rpath.py
@@ -1,0 +1,50 @@
+"""CUDA extensions carry an $ORIGIN DT_RUNPATH so the LD_PRELOAD hook finds the pip CUDA runtime."""
+
+import importlib.util
+from pathlib import Path
+
+import setuptools
+
+
+def _setup_ext_modules(monkeypatch, platform, cuda_major=None):
+    monkeypatch.setenv("TMS_PLATFORM", platform)
+    monkeypatch.delenv("TMS_CUDA_MAJOR", raising=False)
+    if cuda_major is not None:
+        monkeypatch.setenv("TMS_CUDA_MAJOR", cuda_major)
+    captured = {}
+    monkeypatch.setattr(setuptools, "setup", lambda **kwargs: captured.update(kwargs))
+    spec = importlib.util.spec_from_file_location(
+        f"tms_setup_{platform}_{cuda_major}", Path(__file__).parents[1] / "setup.py"
+    )
+    spec.loader.exec_module(importlib.util.module_from_spec(spec))
+    return captured["ext_modules"]
+
+
+def _rpath_args(runtime_rel):
+    return [
+        "-Wl,--enable-new-dtags",
+        f"-Wl,-rpath,$ORIGIN/{runtime_rel}",
+        f"-Wl,-rpath,$ORIGIN/../{runtime_rel}",
+    ]
+
+
+def test_cu13_extensions_carry_pip_runtime_rpath(monkeypatch):
+    ext_modules = _setup_ext_modules(monkeypatch, "cuda", cuda_major="13")
+    assert {ext.name for ext in ext_modules} == {
+        "torch_memory_saver_hook_mode_preload_cu13",
+        "torch_memory_saver_hook_mode_torch_cu13",
+    }
+    for ext in ext_modules:
+        assert ext.extra_link_args == _rpath_args("nvidia/cu13/lib"), ext.name
+
+
+def test_cu12_extensions_carry_component_wheel_rpath(monkeypatch):
+    ext_modules = _setup_ext_modules(monkeypatch, "cuda", cuda_major="12")
+    for ext in ext_modules:
+        assert ext.extra_link_args == _rpath_args("nvidia/cuda_runtime/lib"), ext.name
+
+
+def test_hip_extensions_are_untouched(monkeypatch):
+    ext_modules = _setup_ext_modules(monkeypatch, "hip")
+    for ext in ext_modules:
+        assert ext.extra_link_args == [], ext.name
--- a/torch_memory_saver/hooks/mode_preload.py
+++ b/torch_memory_saver/hooks/mode_preload.py
@@ -1,6 +1,6 @@
 import logging
 import os
-from contextlib import contextmanager
+from contextlib import contextmanager, nullcontext
 from torch_memory_saver.hooks.base import HookUtilBase
 from torch_memory_saver.utils import get_binary_path_from_package, change_env
 
@@ -30,5 +30,24 @@ def configure_subprocess():
 
     new_preload = f"{lib_path}:{current_preload}" if current_preload else lib_path
 
+    # the hook's $ORIGIN RUNPATH only covers co-located pip installs, so also expose the parent's libcudart dir
+    cudart_dir = _mapped_cudart_dir()
+    current_lib = os.environ.get("LD_LIBRARY_PATH", "")
+    new_lib = f"{cudart_dir}:{current_lib}" if current_lib else cudart_dir
+
     with change_env("LD_PRELOAD", new_preload):
-        yield
+        with change_env("LD_LIBRARY_PATH", new_lib) if cudart_dir else nullcontext():
+            yield
+
+
+def _mapped_cudart_dir(maps_path="/proc/self/maps"):
+    try:
+        with open(maps_path) as f:
+            lines = f.read().splitlines()
+    except OSError:
+        return None
+    for line in lines:
+        fields = line.split(maxsplit=5)
+        if len(fields) == 6 and os.path.basename(fields[5]).startswith("libcudart.so."):
+            return os.path.dirname(fields[5])
+    return None
--- a/test/test_configure_subprocess.py
+++ b/test/test_configure_subprocess.py
@@ -51,3 +51,38 @@ def test_restores_ld_preload_after_context():
             with configure_subprocess():
                 assert os.environ["LD_PRELOAD"] != original
         assert os.environ.get("LD_PRELOAD") == original
+
+
+def test_prepends_mapped_cudart_dir_and_restores(monkeypatch):
+    """Split site-packages layout: the child must see the parent's libcudart dir first on LD_LIBRARY_PATH."""
+    from torch_memory_saver.hooks import mode_preload
+
+    monkeypatch.setattr(mode_preload, "_mapped_cudart_dir", lambda: "/site_b/nvidia/cu13/lib")
+    monkeypatch.setenv("LD_LIBRARY_PATH", "/existing/lib")
+    with patch("torch_memory_saver.hooks.mode_preload.get_binary_path_from_package", return_value=FAKE_LIB):
+        with mode_preload.configure_subprocess():
+            assert os.environ["LD_LIBRARY_PATH"] == "/site_b/nvidia/cu13/lib:/existing/lib"
+    assert os.environ["LD_LIBRARY_PATH"] == "/existing/lib"
+
+
+def test_no_mapped_cudart_leaves_ld_library_path_untouched(monkeypatch):
+    from torch_memory_saver.hooks import mode_preload
+
+    monkeypatch.setattr(mode_preload, "_mapped_cudart_dir", lambda: None)
+    monkeypatch.delenv("LD_LIBRARY_PATH", raising=False)
+    with patch("torch_memory_saver.hooks.mode_preload.get_binary_path_from_package", return_value=FAKE_LIB):
+        with mode_preload.configure_subprocess():
+            assert "LD_LIBRARY_PATH" not in os.environ
+
+
+def test_mapped_cudart_dir_parses_proc_maps(tmp_path):
+    from torch_memory_saver.hooks.mode_preload import _mapped_cudart_dir
+
+    maps = tmp_path / "maps"
+    maps.write_text(
+        "aaaa0000-aaaa1000 r-xp 00000000 08:01 1 /usr/lib/libc.so.6\n"
+        "aaaa2000-aaaa3000 rw-p 00000000 00:00 0\n"
+        "aaaa4000-aaaa5000 r-xp 00000000 08:01 2 /site_b/nvidia/cu13/lib/libcudart.so.13\n"
+    )
+    assert _mapped_cudart_dir(str(maps)) == "/site_b/nvidia/cu13/lib"
+    assert _mapped_cudart_dir(str(tmp_path / "absent")) is None
--- a/test/test_preload_hook_loader.py
+++ b/test/test_preload_hook_loader.py
@@ -1,0 +1,225 @@
+"""The shipped CUDA preload hook must resolve libcudart when LD_PRELOADed into a fresh process.
+
+Linux only; needs the installed package and a loadable libcudart.so.<major> (e.g. torch installed).
+"""
+
+import ctypes
+import os
+import re
+import shutil
+import struct
+import subprocess
+import sys
+from pathlib import Path
+from unittest.mock import patch
+
+import pytest
+
+import torch_memory_saver
+from torch_memory_saver import utils
+from torch_memory_saver.hooks import mode_preload
+from torch_memory_saver.hooks.mode_preload import configure_subprocess
+
+_DT_NEEDED, _DT_RPATH, _DT_RUNPATH = 1, 15, 29
+_SHT_DYNAMIC = 6
+_PACKAGE_DIR = Path(torch_memory_saver.__file__).parent
+
+_CHILD = "\n".join([
+    "paths = []",
+    "for line in open('/proc/self/maps'):",
+    "    fields = line.split(maxsplit=5)",
+    "    if len(fields) == 6 and ('libcudart.so' in fields[5] or 'torch_memory_saver_hook' in fields[5]):",
+    "        paths.append(fields[5].strip())",
+    "print('\\n'.join(dict.fromkeys(paths)))",
+])
+
+
+def _runtime_rel(major):
+    return f"nvidia/cu{major}/lib" if major >= 13 else "nvidia/cuda_runtime/lib"
+
+
+def _shipped_cuda_hooks():
+    """(path, CUDA major) of every installed CUDA hook, unsuffixed CUDA 12 compatibility copies included."""
+    hooks = {}
+    for root in (_PACKAGE_DIR, _PACKAGE_DIR.parent):
+        for path in root.glob("torch_memory_saver_hook_mode_*.abi3.so"):
+            hooks.setdefault(path.name, path)
+    if not any("_cu" in name for name in hooks):
+        return []
+    result = []
+    for name, path in sorted(hooks.items()):
+        match = re.search(r"_cu(\d+)\.abi3\.so$", name)
+        result.append((path, int(match.group(1)) if match else 12))
+    return result
+
+
+_CUDA_HOOKS = _shipped_cuda_hooks()
+
+pytestmark = [
+    pytest.mark.skipif(sys.platform != "linux", reason="LD_PRELOAD loader test needs Linux"),
+    pytest.mark.skipif(not _CUDA_HOOKS, reason="no CUDA hooks installed (ROCm/XPU build)"),
+]
+
+
+def _dynamic_section(data):
+    """[(file offset, tag, value)] of the .dynamic entries plus the .dynstr bytes of a little-endian ELF64 object."""
+    assert data[:6] == b"\x7fELF\x02\x01", "expected a little-endian ELF64 object"
+    (shoff,) = struct.unpack_from("<Q", data, 40)
+    shentsize, shnum = struct.unpack_from("<HH", data, 58)
+    headers = [struct.unpack_from("<IIQQQQIIQQ", data, shoff + i * shentsize) for i in range(shnum)]
+    dynamic = next(h for h in headers if h[1] == _SHT_DYNAMIC)
+    strtab = headers[dynamic[6]]
+    entries = []
+    for offset in range(dynamic[4], dynamic[4] + dynamic[5], 16):
+        tag, value = struct.unpack_from("<qQ", data, offset)
+        if tag == 0:
+            break
+        entries.append((offset, tag, value))
+    return entries, data[strtab[4]:strtab[4] + strtab[5]]
+
+
+def _dynamic_strings(path, wanted_tag):
+    entries, strings = _dynamic_section(path.read_bytes())
+    return [strings[value:strings.index(b"\0", value)].decode() for _, tag, value in entries if tag == wanted_tag]
+
+
+def _empty_runpath(path):
+    """Point DT_RUNPATH/DT_RPATH at the empty string, leaving every other byte intact."""
+    data = bytearray(path.read_bytes())
+    entries, _ = _dynamic_section(bytes(data))
+    for offset, tag, _ in entries:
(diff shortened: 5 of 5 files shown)


fzyzcjy:
Great job! could you please remove the comments and cleanup the code a bit

rishabhsinha17:
Done: comments removed, code tightened, test still fails without / passes with the fix.

setup.py
```diff
@@ -162,6 +163,9 @@ def _create_ext_modules(platform):
                 "(use `make build-wheel-multi-cuda` or scripts/build_multi_cuda.sh)."
             )
         name_suffix = f"_cu{cuda_major}"
+        # RPATH so the LD_PRELOAD hook finds the pip-installed CUDA runtime.
+        runtime_rel = f"nvidia/cu{cuda_major}/lib" if int(cuda_major) >= 13 else "nvidia/cuda_runtime/lib"
+        extra_link_args = [f"-Wl,-rpath,$ORIGIN/{runtime_rel}", f"-Wl,-rpath,$ORIGIN/../{runtime_rel}"]
```
fzyzcjy: **P1 — This still crashes when the packages live in different site-packages roots.** These RUNPATH entries are anchored only to the hook DSO, but `torch_memory_saver` does not depend on the NVIDIA runtime package and Python can legitimately load torch/CUDA from system site-packages while TMS is installed in a user site or a `--system-site-packages` venv. In that layout `libcudart.so` is under the torch/runtime package root, neither of these directories exists relative to the hook, and the `LD_PRELOAD` load still exits before Python can discover torch. Please either derive/inject the actual runtime directory at `configure_subprocess()` time (the canonical place where Python can inspect torch/NVIDIA), or define and enforce a co-location dependency/contract. Add a split-site-packages preload subprocess regression test; the current tests only compare linker strings and cannot catch this failure.

fzyzcjy (changes requested):
(by agent, ignore this)

I found one loader-path correctness issue that still leaves a supported Python installation layout broken. The co-located pip layout looks correct, and I did not find lifecycle or performance regressions there.

fzyzcjy (commented):
hi @rishabhsinha17 the code generally lgtm, could you please

1. ping your agent to enhance tests (e.g. my agent made some suggestsions as follows)
2. run all tests on your machine and paste evidence all tests pass

---

The regression test needs to validate the shipped artifact and the original loader failure, not only the `Extension` configuration.

test/test_setup_rpath.py
```diff
@@ -0,0 +1,46 @@
...
+    return captured["ext_modules"]
+
+
+def _rpath_args(runtime_rel):
+    return [f"-Wl,-rpath,$ORIGIN/{runtime_rel}", f"-Wl,-rpath,$ORIGIN/../{runtime_rel}"]
+
+
+def test_cu13_extensions_carry_pip_runtime_rpath(monkeypatch):
+    ext_modules = _setup_ext_modules(monkeypatch, "cuda", cuda_major="13")
+    assert {ext.name for ext in ext_modules} == {
```
fzyzcjy, line 33: **Please strengthen this into an artifact-level loader regression test.** The current assertion duplicates the strings from `setup.py`, so it stays green if setuptools/linker quoting is wrong, wheel repair or merge drops the dynamic tag, or the resulting preload object still cannot resolve `libcudart`. The coverage should exercise the complete release path:

- build the CUDA 12 and CUDA 13 wheels and run the normal multi-CUDA merge step;
- inspect every shipped CUDA hook with `readelf -d` (including the unsuffixed CUDA 12 compatibility copies) and assert the final ELF contains the expected `DT_RUNPATH`;
- install the final wheel together with the matching official NVIDIA runtime wheel in a clean container where `ldconfig` cannot already resolve `libcudart`;
- unset `LD_LIBRARY_PATH`, set `LD_PRELOAD` to the installed preload hook, and start a fresh Python subprocess for both CUDA 12 and CUDA 13; assert it reaches Python and exits successfully;
- include a negative control built without these flags and assert that the same subprocess fails to start, so the test proves this RUNPATH is what fixes the reported failure.

The lightweight configuration assertions can remain as fast unit tests, but they are not sufficient regression coverage for this packaging/loader fix.

setup.py
rishabhsinha17: Fixed in 0c00925 with both layers, since each covers what the other cannot:

1. DT_RUNPATH in the hook (`-Wl,--enable-new-dtags -Wl,-rpath,$ORIGIN/<rel> -Wl,-rpath,$ORIGIN/../<rel>`). Covers the pip layout where the wheel and the NVIDIA runtime wheel share a site-packages root with no env changes at all: `LD_PRELOAD=<hook> python` alone starts. RUNPATH rather than RPATH so an LD_LIBRARY_PATH entry still takes precedence over it.
2. `configure_subprocess()` now prepends the directory of the parent's `libcudart.so.<major>` to the child's `LD_LIBRARY_PATH` (`utils.find_cuda_runtime_dir`): the copy already mapped into the parent wins (`/proc/self/maps`, i.e. the one torch loaded), else the pip runtime wheel found through the `nvidia` namespace package (`nvidia/cu13/lib` or `nvidia/cuda_runtime/lib`). This is the split layout you describe (TMS in a user site or `--system-site-packages` venv, runtime in the system site-packages), and also system or conda CUDA where no `$ORIGIN` entry exists. The major comes from the selected hook's `_cu<major>` suffix, so ROCm hooks are untouched; no-op when the directory is already listed or nothing is found. Same mechanism sglang applies at its layer in sgl-project/sglang#36643, so the two stay consistent.

Regression tests: `test_configure_subprocess.py::test_split_site_packages_prepends_runtime_dir_to_ld_library_path[12|13]` (fake roots A/B, asserts the child env and the restore), and the artifact-level `test_preload_hook_loader.py::test_split_site_packages_needs_runtime_dir_injection`, which copies the installed hook into a root with no `nvidia/` sibling, shows the LD_PRELOAD child dies with `libcudart.so.<major>: cannot open shared object file` on RUNPATH alone, then passes through `configure_subprocess()` with the child mapping the injected `libcudart`. Evidence in the top-level comment.

rishabhsinha17:
Pushed 0c00925 (single amended commit). Changes since the last review:

- Two layers: the `$ORIGIN` DT_RUNPATH stays (now with `--enable-new-dtags`, so it is RUNPATH on every toolchain and LD_LIBRARY_PATH keeps precedence), and `configure_subprocess()` additionally prepends the parent's `libcudart.so.<major>` directory to the child's `LD_LIBRARY_PATH` (mapped copy from `/proc/self/maps` first, else the pip `nvidia/*/lib` wheel). Details in the P1 thread.
- Tests: `test/test_preload_hook_loader.py` runs against the installed wheel on Linux: DT_RUNPATH of every shipped CUDA hook (cu12, cu13, unsuffixed compat copies), the co-located LD_PRELOAD subprocess with no LD_LIBRARY_PATH, the split-site-packages subprocess (fails on RUNPATH alone, passes via `configure_subprocess()`), and a negative control with the RUNPATH emptied in place. Plus unit tests for the injection and discovery helpers. This supersedes the "did not rebuild a cu13 wheel" note in the description.

Platform, stated plainly: everything below ran on Linux aarch64 (Ubuntu 24.04 VM, g++ 13.3.0, GNU ld 2.42, apt `cuda-cudart-dev`/`cuda-driver-dev` 12.8 and 13.0, no GPU). Builds used the repo's own steps (`setup.py bdist_wheel --py-limited-api cp39` with setuptools 75.0.0 / wheel 0.41.0 as in `scripts/build_in_docker.sh`, `rename_wheels.sh`, `merge_cuda_wheels.py`), not the `pytorch/manylinuxaarch64-builder` images (8 GB each, no disk for them here). x86_64 was not exercised, and the GPU suite (`test_examples.py`) was not run: no NVIDIA GPU on this side.

### readelf on the built wheel (branch, every shipped hook)

```text
$ readelf -d torch_memory_saver_hook_mode_preload.abi3.so | grep -E 'NEEDED.*(cuda|cudart)|RUNPATH|RPATH'
 0x0000000000000001 (NEEDED)             Shared library: [libcuda.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libcudart.so.12]
 0x000000000000001d (RUNPATH)            Library runpath: [$ORIGIN/nvidia/cuda_runtime/lib:$ORIGIN/../nvidia/cuda_runtime/lib]
$ readelf -d torch_memory_saver_hook_mode_preload_cu12.abi3.so | grep -E 'NEEDED.*(cuda|cudart)|RUNPATH|RPATH'
 0x0000000000000001 (NEEDED)             Shared library: [libcuda.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libcudart.so.12]
 0x000000000000001d (RUNPATH)            Library runpath: [$ORIGIN/nvidia/cuda_runtime/lib:$ORIGIN/../nvidia/cuda_runtime/lib]
$ readelf -d torch_memory_saver_hook_mode_preload_cu13.abi3.so | grep -E 'NEEDED.*(cuda|cudart)|RUNPATH|RPATH'
 0x0000000000000001 (NEEDED)             Shared library: [libcuda.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libcudart.so.13]
 0x000000000000001d (RUNPATH)            Library runpath: [$ORIGIN/nvidia/cu13/lib:$ORIGIN/../nvidia/cu13/lib]
$ readelf -d torch_memory_saver_hook_mode_torch.abi3.so | grep -E 'NEEDED.*(cuda|cudart)|RUNPATH|RPATH'
 0x0000000000000001 (NEEDED)             Shared library: [libcuda.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libcudart.so.12]
 0x000000000000001d (RUNPATH)            Library runpath: [$ORIGIN/nvidia/cuda_runtime/lib:$ORIGIN/../nvidia/cuda_runtime/lib]
$ readelf -d torch_memory_saver_hook_mode_torch_cu12.abi3.so | grep -E 'NEEDED.*(cuda|cudart)|RUNPATH|RPATH'
 0x0000000000000001 (NEEDED)             Shared library: [libcuda.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libcudart.so.12]
 0x000000000000001d (RUNPATH)            Library runpath: [$ORIGIN/nvidia/cuda_runtime/lib:$ORIGIN/../nvidia/cuda_runtime/lib]
$ readelf -d torch_memory_saver_hook_mode_torch_cu13.abi3.so | grep -E 'NEEDED.*(cuda|cudart)|RUNPATH|RPATH'
 0x0000000000000001 (NEEDED)             Shared library: [libcuda.so.1]
 0x0000000000000001 (NEEDED)             Shared library: [libcudart.so.13]
 0x000000000000001d (RUNPATH)            Library runpath: [$ORIGIN/nvidia/cu13/lib:$ORIGIN/../nvidia/cu13/lib]
```

Control wheel built from origin/master with the same steps: all six hooks have the same NEEDED lines and no RUNPATH/RPATH entry. Both wheels pass `release_checks.validate_wheel` (binary set, ELF machine, cu12 compatibility copies byte-identical, RECORD, tag); `cmp` confirms the unsuffixed copies equal the cu12 ones.

Linker invocation as recorded by setuptools, showing `$ORIGIN` reaches ld unexpanded:

```text
aarch64-linux-gnu-g++ ... -lcuda -lcudart -o build/lib.linux-aarch64-cpython-312/torch_memory_saver_hook_mode_preload_cu13.abi3.so -Wl,--enable-new-dtags -Wl,-rpath,$ORIGIN/nvidia/cu13/lib -Wl,-rpath,$ORIGIN/../nvidia/cu13/lib
```

### Clean container, LD_PRELOAD subprocess

`python:3.13-slim` (Debian 13, aarch64), CPU torch 2.14.0, `nvidia-cuda-runtime-cu12==12.8.90` and `nvidia-cuda-runtime==13.0.96` (the official runtime wheels), the merged wheel installed with `--no-deps` into the same site-packages, and the CUDA driver stub copied to `/usr/local/lib/libcuda.so.1` so the hook's other NEEDED resolves without a GPU. Loader default search before the tests:

```text
$ ldconfig -p | grep -E "libcuda\.so|libcudart"
	libcuda.so.1 (libc6,AArch64) => /usr/local/lib/libcuda.so.1
```

Direct check, LD_PRELOAD only, `LD_LIBRARY_PATH` unset:

```text
branch cu13: python started under LD_PRELOAD   exit=0
branch cu12: python started under LD_PRELOAD   exit=0
master cu13: python: error while loading shared libraries: libcudart.so.13: cannot open shared object file: No such file or directory   exit=127
master cu12: python: error while loading shared libraries: libcudart.so.12: cannot open shared object file: No such file or directory   exit=127
```

`test/test_preload_hook_loader.py` with the branch wheel. The pytest process gets `LD_LIBRARY_PATH=<runtime dir>` (same as the release harness does), which selects the major; every child is started with libcudart directories stripped from `LD_LIBRARY_PATH`.

```text
LD_LIBRARY_PATH=$SITE/nvidia/cu13/lib python -m pytest test/test_preload_hook_loader.py -v
test/test_preload_hook_loader.py::test_shipped_cuda_hook_carries_origin_runpath[torch_memory_saver_hook_mode_preload.abi3.so] PASSED
test/test_preload_hook_loader.py::test_shipped_cuda_hook_carries_origin_runpath[torch_memory_saver_hook_mode_preload_cu12.abi3.so] PASSED
test/test_preload_hook_loader.py::test_shipped_cuda_hook_carries_origin_runpath[torch_memory_saver_hook_mode_preload_cu13.abi3.so] PASSED
test/test_preload_hook_loader.py::test_shipped_cuda_hook_carries_origin_runpath[torch_memory_saver_hook_mode_torch.abi3.so] PASSED
test/test_preload_hook_loader.py::test_shipped_cuda_hook_carries_origin_runpath[torch_memory_saver_hook_mode_torch_cu12.abi3.so] PASSED
test/test_preload_hook_loader.py::test_shipped_cuda_hook_carries_origin_runpath[torch_memory_saver_hook_mode_torch_cu13.abi3.so] PASSED
test/test_preload_hook_loader.py::test_colocated_hook_preloads_without_ld_library_path PASSED
test/test_preload_hook_loader.py::test_split_site_packages_needs_runtime_dir_injection PASSED
test/test_preload_hook_loader.py::test_hook_with_emptied_runpath_cannot_preload PASSED
============================== 9 passed in 0.71s ===============================

LD_LIBRARY_PATH=$SITE/nvidia/cuda_runtime/lib python -m pytest test/test_preload_hook_loader.py -v
(same nine cases, cu12 hook under test)
============================== 9 passed in 0.67s ===============================
```

Same module against the origin/master wheel (`pip install --force-reinstall`):

```text
test_shipped_cuda_hook_carries_origin_runpath[...] FAILED x6   (assert [] == ['$ORIGIN/nvidia/cu13/lib:$ORIGIN/../nvidia/cu13/lib'])
test_colocated_hook_preloads_without_ld_library_path FAILED    (/usr/local/bin/python: error while loading shared libraries: libcudart.so.13: cannot open shared object file: No such file or directory, assert 127 == 0)
test_split_site_packages_needs_runtime_dir_injection ERROR     (module 'torch_memory_saver.utils' has no attribute 'find_cuda_runtime_dir')
test_hook_with_emptied_runpath_cannot_preload ERROR            (same)
========================= 7 failed, 2 errors in 0.71s ==========================
```

The negative control inside the passing run: the co-located copy of the hook loads (`libcudart.so.13` mapped from the wheel dir), then with DT_RUNPATH pointed at the empty string (`readelf` would show `Library runpath: []`, every other byte identical) the same child fails with `libcudart.so.13: cannot open shared object file`. Done in-test with a 30-line ELF64 .dynamic rewrite rather than `patchelf` so the release containers need no extra dependency.

### Non-GPU suite on my machine (macOS arm64, Python 3.13.3, torch 2.14.0 CPU)

```text
$ PYTHONPATH=. python -m pytest test/test_configure_subprocess.py test/test_setup_rpath.py test/test_utils.py test/test_merge_cuda_wheels.py test/test_preload_hook_loader.py -v
test/test_configure_subprocess.py::test_sets_ld_preload_when_unset PASSED
test/test_configure_subprocess.py::test_prepends_without_overwriting_existing PASSED
test/test_configure_subprocess.py::test_restores_ld_preload_after_context PASSED
test/test_configure_subprocess.py::test_split_site_packages_prepends_runtime_dir_to_ld_library_path[12] PASSED
test/test_configure_subprocess.py::test_split_site_packages_prepends_runtime_dir_to_ld_library_path[13] PASSED
test/test_configure_subprocess.py::test_mapped_runtime_wins_over_wheel_lookup PASSED
test/test_configure_subprocess.py::test_runtime_dir_already_listed_is_not_duplicated PASSED
test/test_configure_subprocess.py::test_unknown_runtime_dir_leaves_ld_library_path_unset PASSED
test/test_configure_subprocess.py::test_unsuffixed_rocm_hook_leaves_ld_library_path_alone PASSED
test/test_setup_rpath.py::test_cu13_extensions_carry_pip_runtime_rpath PASSED
test/test_setup_rpath.py::test_cu12_extensions_carry_component_wheel_rpath PASSED
test/test_setup_rpath.py::test_hip_extensions_are_untouched PASSED
test/test_utils.py::test_get_binary_path_from_package_uses_unsuffixed_rocm_binary PASSED
test/test_utils.py::test_get_binary_path_from_package_keeps_cuda_suffix_selection PASSED
test/test_utils.py::test_change_env_leaves_absent_key_absent PASSED
test/test_utils.py::test_change_env_restores_previous_value PASSED
test/test_utils.py::test_cpu_backup_backend_requires_enable PASSED
test/test_utils.py::test_cpu_backup_backend_rejected_on_xpu PASSED
test/test_utils.py::test_cpu_backup_backend_rejects_mmap_on_rocm PASSED
test/test_utils.py::test_mapped_library_dir_reads_proc_maps PASSED
test/test_utils.py::test_nvidia_wheel_library_dir_covers_both_pip_layouts PASSED
test/test_merge_cuda_wheels.py::test_merge_rewrites_internal_wheel_tag_to_match_output_filename PASSED
test/test_merge_cuda_wheels.py::test_merge_duplicates_cuda12_binaries_under_compatibility_names PASSED
test/test_preload_hook_loader.py::test_shipped_cuda_hook_carries_origin_runpath[NOTSET] SKIPPED
test/test_preload_hook_loader.py::test_colocated_hook_preloads_without_ld_library_path SKIPPED
test/test_preload_hook_loader.py::test_split_site_packages_needs_runtime_dir_injection SKIPPED
test/test_preload_hook_loader.py::test_hook_with_emptied_runpath_cannot_preload SKIPPED
======================== 23 passed, 4 skipped in 0.75s =========================
```

The four skips are the Linux-only module. Same test files against origin/master product code: `9 failed, 12 passed` (the two setup.py capture tests, the five new configure_subprocess tests, the two discovery helper tests).

Two notes for the release harness: `test_preload_hook_loader.py` adds no skips in a CUDA container as long as `ldconfig` does not already resolve `libcudart.so.<major>`; if a cell's image registers a system libcudart, the three loader tests skip with reason `ldconfig already resolves libcudart.so.<major>` and would need a matrix entry. And the tests strip only LD_LIBRARY_PATH directories that contain `libcudart.so*`, so a Lupine shim directory holding only `libcuda.so.1` is kept.

torch_memory_saver/hooks/mode_preload.py
fzyzcjy, line 1: would be great to avoid lengthy comments, also iirc last time this file (and uttils.py) is not changed and would be great to keep things simple - only need to handle https://github.com/fzyzcjy/torch_memory_saver/pull/108#discussion_r3885766180 at the old short code

rishabhsinha17: Done in a9403e9. utils.py, README.md and test_utils.py are back to master untouched. The code change is setup.py +9 (the RUNPATH link args) and mode_preload.py +21/-2: configure_subprocess() reads the parent's libcudart dir from /proc/self/maps and prepends it to LD_LIBRARY_PATH, one short comment each. PR total is +340/-2, of which +310 is the coverage from the earlier test thread (artifact-level loader tests with negative control, plus small unit tests). Also rebased on current master.

I read your note as: keep the runtime dir handling but inline it at the old configure_subprocess() site, no new helpers in utils.py. So the nvidia wheel fallback and the hook suffix version match are gone; the mapped libcudart is the copy torch loaded, and when nothing is mapped it is a no-op and the RUNPATH covers the co-located pip layout.

Local runs (macOS, the Linux-only loader tests skip): test_configure_subprocess + test_setup_rpath + test_utils + test_preload_hook_loader give 16 passed, 4 skipped. With setup.py and mode_preload.py reverted to master the new tests fail (5 failed), restored they pass. Loader test assertions are unchanged from the Linux evidence pasted on Sep 2; only the runtime_dir fixture changed, it now derives from /proc/self/maps.

fzyzcjy (approved):

fzyzcjy:
Ordinary CUDA 12 regression run on b5b027a6a727ebf6f2e7cd0de54668efa503746a: **39 passed, 14 skipped in 49.44s**, exit code 0.

- Environment: NVIDIA GeForce RTX 4090 D, driver 580.173.02, Python 3.11.11, PyTorch 2.6.0+cu124, CUDA 12.4; GPU 0 only.
- Fresh container: `pytorch/pytorch:2.6.0-cuda12.4-cudnn9-devel`, image ID `sha256:7d57e307bd9c152da255651dc3458beaecf35fcdf4752fbbc14217b9be0307aa`.
- Input: clean `git archive` of the commit above, extracted into an isolated directory. Extensions were rebuilt from that source; this was a source-checkout run, not final-wheel validation.
- Dependencies: pytest 8.3.5 and nvidia-ml-py 12.570.86.

Commands inside the container:

```bash
cd /workspace
export CUDA_VISIBLE_DEVICES=0
TMS_CUDA_MAJOR=12 uv run --no-project --python /opt/conda/bin/python python setup.py build_ext --inplace
readelf -d /workspace/torch_memory_saver_hook_mode_preload_cu12.abi3.so
uv run --no-project --python /opt/conda/bin/python python -m pytest test -vv -ra
```

- Passed: mapped CUDART path parsing, environment-variable prepending/restoration, CUDA 12/13 link-argument configuration, unchanged HIP link configuration, ELF RUNPATH assertions for both rebuilt CUDA 12 hooks, and the applicable ordinary preload/torch scenarios.
- `readelf` confirmed `DT_RUNPATH` is `$ORIGIN/nvidia/cuda_runtime/lib:$ORIGIN/../nvidia/cuda_runtime/lib`.
- Skips: 5 multi-device cases (only one GPU exposed), 6 XPU-only cases, and the 3 loader integration cases below. All three loader cases skipped because `ldconfig already resolves libcudart.so.12`:
  - `test_colocated_hook_preloads_without_ld_library_path`
  - `test_split_site_packages_needs_runtime_dir_injection`
  - `test_hook_with_emptied_runpath_cannot_preload`
- Coverage limit: this run does **not** demonstrate those three missing-runtime-search-path scenarios. CUDA 13 runtime, ARM64, and the full wheel matrix were not run; CUDA 13 coverage here is configuration-level only.
- The temporary test container was removed successfully. No source changes were needed.

